PR #2, opened 2026-04-11
feat: staking operations (add, remove, list)

moorkh:
## Summary

- **`btt stake list`** -- read-only query of all stakes for a coldkey address (via `--wallet` or `--ss58`). Uses `StakingHotkeys` storage to discover hotkeys, then queries `Alpha` triple map for per-subnet stake amounts, with `Stake` double map fallback.
- **`btt stake add`** -- stake TAO from coldkey to hotkey on a specific subnet. Constructs and submits `SubtensorModule::add_stake(hotkey, netuid, amount_rao)` extrinsic signed by the decrypted coldkey.
- **`btt stake remove`** -- unstake TAO. Supports `--amount` or `--all` flag which queries current stake before submitting `SubtensorModule::remove_stake`.
- Adds `wallet_keys` module with coldkey decryption (NaCl/scrypt encrypted or plaintext JSON), TAO/RAO conversion, key generation, signing, and verification.
- Adds CLI definitions for `StakeAction` enum and wires routing in `main.rs`.
- Adds dependencies: `rpassword`, `xsalsa20poly1305`, `scrypt`, `zeroize`, `rand`.

## Key design decisions

- TAO amounts accepted as `f64`, converted to RAO (`u64`, 1 TAO = 1e9 RAO) with validation for NaN/Inf/negative/overflow.
- Coldkey decryption tries plaintext JSON first, then prompts for password and decrypts via scrypt + xsalsa20poly1305 (btcli-compatible format).
- Key material is zeroized after use.
- All extrinsic submissions use `sign_and_submit_then_watch_default` and wait for finalization.

## Test plan

- [x] `cargo build` -- zero errors
- [x] `cargo clippy -- -D warnings` -- zero warnings
- [x] `cargo test` -- 49 tests pass, including:
  - TAO/RAO conversion (whole, fractional, zero, large, negative, NaN, Infinity)
  - RAO-to-TAO string formatting (whole, fractional, zero, small, precise)
  - SS58 address resolution and validation
  - Encrypt/decrypt roundtrip, wrong-password rejection
  - Wallet creation roundtrip, mnemonic/seed recovery
  - Sign/verify roundtrip
- [ ] Manual test against testnet: `btt stake list --ss58 <addr> --network test`
- [ ] Manual test: `btt stake add --wallet <name> --hotkey <ss58> --netuid 1 --amount 0.1 --network test`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -958,6 +958,10 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "90dbd31c98227229239363921e60fcf5e558e43ec69094d46fc4996f08d1d5bc"
 dependencies = [
  "bitcoin_hashes 0.14.1",
+ "rand 0.8.5",
+ "rand_core 0.6.4",
+ "serde",
+ "unicode-normalization",
 ]
 
 [[package]]
@@ -1378,12 +1382,19 @@ dependencies = [
 name = "btt"
 version = "0.1.0"
 dependencies = [
+ "bip39",
  "clap",
+ "hex",
+ "rand 0.8.5",
+ "rpassword",
+ "scrypt",
  "serde",
  "serde_json",
  "sp-core",
  "subxt",
  "tokio",
+ "xsalsa20poly1305",
+ "zeroize",
 ]
 
 [[package]]
@@ -1509,6 +1520,7 @@ checksum = "773f3b9af64447d2ce9850330c473515014aa235e6a783b02db81ff39e4a3dad"
 dependencies = [
  "crypto-common",
  "inout",
+ "zeroize",
 ]
 
 [[package]]
@@ -6748,6 +6760,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f8ed6a7761f76e3b9f92dfb0a60a6a6477c61024b775147ff0973a02653abaf2"
 dependencies = [
  "digest 0.10.7",
+ "hmac 0.12.1",
  "password-hash",
 ]
 
@@ -7956,6 +7969,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "rpassword"
+version = "5.0.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ffc936cf8a7ea60c58f030fd36a612a48f440610214dc54bc36431f9ea0c3efb"
+dependencies = [
+ "libc",
+ "winapi",
+]
+
+[[package]]
 name = "ruint"
 version = "1.17.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -8185,6 +8208,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "96b02de82ddbe1b636e6170c21be622223aea188ef2e139be0a5b219ec215323"
 dependencies = [
  "bytemuck",
+]
+
+[[package]]
+name = "salsa20"
+version = "0.10.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "97a22f5af31f73a954c10289c93e8a50cc23d971e80ee446f1f6f7137a088213"
+dependencies = [
+ "cipher",
 ]
 
 [[package]]
@@ -8463,6 +8495,18 @@ name = "scratch"
 version = "1.0.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d68f2ec51b097e4c1a75b681a8bec621909b5e91f15bb7b840c4f2f7b01148b2"
+
+[[package]]
+name = "scrypt"
+version = "0.11.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0516a385866c09368f0b5bcd1caff3366aace790fcd46e2bb032697bb172fd1f"
+dependencies = [
+ "password-hash",
+ "pbkdf2",
+ "salsa20",
+ "sha2 0.10.9",
+]
 
 [[package]]
 name = "sec1"
@@ -11940,6 +11984,19 @@ dependencies = [
  "staging-xcm",
  "staging-xcm-builder",
  "staging-xcm-executor",
+]
+
+[[package]]
+name = "xsalsa20poly1305"
+version = "0.9.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "02a6dad357567f81cd78ee75f7c61f1b30bb2fe4390be8fb7c69e2ac8dffb6c7"
+dependencies = [
+ "aead",
+ "poly1305",
+ "salsa20",
+ "subtle",
+ "zeroize",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,4 +11,11 @@ tokio = { version = "1", features = ["full"] }
 clap = { version = "4", features = ["derive"] }
 serde = { version = "1", features = ["derive"] }
 serde_json = "1"
-sp-core = { version = "34", default-features = false }
+sp-core = { version = "34", default-features = false, features = ["std"] }
+bip39 = { version = "2", features = ["rand"] }
+rpassword = "5"
+zeroize = { version = "1", features = ["derive"] }
+xsalsa20poly1305 = "0.9"
+scrypt = "0.11"
+rand = "0.8"
+hex = "0.4"
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -37,6 +37,12 @@ pub enum Command {
         action: WalletAction,
     },
 
+    /// Staking operations
+    Stake {
+        #[command(subcommand)]
+        action: StakeAction,
+    },
+
     /// Emit SKILL.md for AI agent integration
     Skill,
 }
@@ -57,4 +63,146 @@ pub enum ChainAction {
 pub enum WalletAction {
     /// List wallets in ~/.bittensor/wallets/
     List,
+
+    /// Create a new wallet (coldkey + hotkey pair)
+    Create {
+        /// Wallet name
+        #[arg(long)]
+        name: String,
+        /// Hotkey name
+        #[arg(long, default_value = "default")]
+        hotkey: String,
+        /// Number of mnemonic words (12 or 24)
+        #[arg(long, default_value_t = 12)]
+        n_words: u32,
+    },
+
+    /// Generate a new coldkey only
+    NewColdkey {
+        /// Wallet name
+        #[arg(long)]
+        name: String,
+        /// Number of mnemonic words (12 or 24)
+        #[arg(long, default_value_t = 12)]
+        n_words: u32,
+    },
+
+    /// Generate a new hotkey for an existing wallet
+    NewHotkey {
+        /// Wallet name
+        #[arg(long)]
+        name: String,
+        /// Hotkey name
+        #[arg(long, default_value = "default")]
+        hotkey: String,
+        /// Number of mnemonic words (12 or 24)
+        #[arg(long, default_value_t = 12)]
+        n_words: u32,
+    },
+
+    /// Restore a coldkey from mnemonic or seed
+    RegenColdkey {
+        /// Wallet name
+        #[arg(long)]
+        name: String,
+        /// BIP39 mnemonic phrase
+        #[arg(long)]
+        mnemonic: Option<String>,
+        /// Hex-encoded seed (0x...)
+        #[arg(long)]
+        seed: Option<String>,
+    },
+
+    /// Restore a hotkey from mnemonic or seed
+    RegenHotkey {
+        /// Wallet name
+        #[arg(long)]
+        name: String,
+        /// Hotkey name
+        #[arg(long, default_value = "default")]
+        hotkey: String,
+        /// BIP39 mnemonic phrase
+        #[arg(long)]
+        mnemonic: Option<String>,
+        /// Hex-encoded seed (0x...)
+        #[arg(long)]
+        seed: Option<String>,
+    },
+
+    /// Sign a message with a wallet key
+    Sign {
+        /// Wallet name
+        #[arg(long)]
+        name: String,
+        /// Hotkey name (when using --use-hotkey)
+        #[arg(long, default_value = "default")]
+        hotkey: String,
+        /// Message to sign
+        #[arg(long)]
+        message: String,
+        /// Sign with hotkey instead of coldkey
+        #[arg(long, default_value_t = false)]
+        use_hotkey: bool,
+    },
+
+    /// Verify a signature
+    Verify {
+        /// Message that was signed
+        #[arg(long)]
+        message: String,
+        /// Hex-encoded signature (0x...)
+        #[arg(long)]
+        signature: String,
+        /// SS58 address of the signer
+        #[arg(long)]
+        ss58: String,
+    },
+}
+
+#[derive(Subcommand, Debug)]
+pub enum StakeAction {
+    /// List all stakes for a wallet
+    List {
+        /// Wallet name (reads coldkeypub.txt for the SS58 address)
+        #[arg(long)]
+        wallet: Option<String>,
+        /// SS58 address to query directly (alternative to --wallet)
+        #[arg(long)]
+        ss58: Option<String>,
+    },
+
+    /// Stake TAO from coldkey to hotkey on a subnet
+    Add {
+        /// Wallet name (coldkey will be decrypted for signing)
+        #[arg(long)]
+        wallet: String,
+        /// Hotkey SS58 address to stake to
+        #[arg(long)]
+        hotkey: String,
+        /// Subnet ID
+        #[arg(long)]
+        netuid: u16,
+        /// Amount in TAO (e.g. 10.5)
+        #[arg(long)]
+        amount: f64,
+    },
+
+    /// Unstake TAO from hotkey back to coldkey
+    Remove {
+        /// Wallet name (coldkey will be decrypted for signing)
+        #[arg(long)]
+        wallet: String,
+        /// Hotkey SS58 address to unstake from
+        #[arg(long)]
+        hotkey: String,
+        /// Subnet ID
+        #[arg(long)]
+        netuid: u16,
+        /// Amount in TAO (e.g. 10.5). Ignored if --all is set.
+        #[arg(long, required_unless_present = "all")]
+        amount: Option<f64>,
+        /// Unstake all TAO from this hotkey on this subnet
+        #[arg(long, default_value_t = false)]
+        all: bool,
+    },
 }
--- a/src/commands/mod.rs
+++ b/src/commands/mod.rs
@@ -1,3 +1,5 @@
 pub mod chain;
 pub mod skill;
+pub mod stake;
 pub mod wallet;
+pub mod wallet_keys;
--- a/src/commands/skill.rs
+++ b/src/commands/skill.rs
@@ -39,6 +39,21 @@ btt chain balance <ss58_address>
 btt wallet list
 ```
 
+### Stake
+
+```bash
+# List all stakes for a wallet
+btt stake list --wallet <name>
+btt stake list --ss58 <address>
+
+# Add stake (TAO from coldkey to hotkey on a subnet)
+btt stake add --wallet <name> --hotkey <ss58> --netuid <u16> --amount <TAO>
+
+# Remove stake (unstake TAO from hotkey back to coldkey)
+btt stake remove --wallet <name> --hotkey <ss58> --netuid <u16> --amount <TAO>
+btt stake remove --wallet <name> --hotkey <ss58> --netuid <u16> --all
+```
+
 ### Skill
 
 ```bash
--- a/src/commands/stake.rs
+++ b/src/commands/stake.rs
@@ -1,0 +1,536 @@
+use std::time::Duration;
+
+use serde::Serialize;
+use sp_core::crypto::Ss58Codec;
+use sp_core::sr25519;
+use sp_core::Pair as PairTrait;
+use subxt::dynamic::{At, Value};
+use subxt::tx::PairSigner;
+use subxt::PolkadotConfig;
+
+use crate::commands::chain::parse_ss58;
+use crate::commands::wallet_keys::{
+    decrypt_coldkey_interactive, rao_to_tao_string, resolve_coldkey_address, tao_to_rao,
+};
+use crate::error::BttError;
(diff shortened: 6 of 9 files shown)


moorkh:
## Testing Plan — PR #2: Staking operations

This plan covers manual integration tests and unit test verification for `btt stake list`, `btt stake add`, and `btt stake remove`.

All write operations (add, remove) must be performed against **testnet only** (`--network test`). Never submit extrinsics against Finney mainnet during testing.

---

### Manual Testing

#### `btt stake list`

- [ ] `btt stake list --ss58 <known_validator_address>` against Finney mainnet — verify JSON output contains `address` and a non-empty `stakes` array. Each entry should have `hotkey`, `netuid`, and `stake_tao` fields. Confirm `stake_tao` values are in TAO (decimal, e.g. `"10.5"`), not raw RAO integers.
- [ ] `btt stake list --ss58 <address_with_no_stakes>` — verify output is `{"ok":true,"data":{"address":"...","stakes":[]}}` with an empty array.
- [ ] `btt stake list --wallet <name>` — verify the command reads `~/.bittensor/wallets/<name>/coldkeypub.txt` for the SS58 address and produces the same output as `--ss58`.
- [ ] `btt stake list --wallet <name> --ss58 <addr>` — expect exit code 1 with error: "provide either --wallet or --ss58, not both".
- [ ] `btt stake list` (no `--wallet` or `--ss58`) — expect exit code 1 with error: "provide either --wallet or --ss58".

#### `btt stake add` (testnet only)

- [ ] `btt stake add --wallet <name> --hotkey <ss58> --netuid 1 --amount 0.1 --network test` — verify password prompt appears with no echo. On correct password, verify extrinsic is submitted and output contains `tx_hash`, `block`, `action: "add_stake"`, `amount_tao: "0.1"`, `hotkey`, and `netuid: 1`.
- [ ] After the above: `btt stake list --ss58 <coldkey_ss58> --network test` — verify the stake appears in the list.

#### `btt stake remove` (testnet only)

- [ ] `btt stake remove --wallet <name> --hotkey <ss58> --netuid 1 --amount 0.1 --network test` — verify password prompt, then successful removal. Output should contain `action: "remove_stake"` and `amount_tao: "0.1"`.
- [ ] `btt stake remove --wallet <name> --hotkey <ss58> --netuid 1 --all --network test` — verify the command queries current stake, then submits removal for the full amount. Confirm `stake list` shows zero stake afterward.
- [ ] `btt stake remove --wallet <name> --hotkey <ss58> --netuid 1` (no `--amount`, no `--all`) — clap should reject this at argument parsing before any network call. Expect an error referencing the missing required argument.

#### Amount edge cases

- [ ] `--amount 0` — expect exit code 1 with `INVALID_INPUT` error: "stake amount must be greater than zero".
- [ ] `--amount -1` — expect exit code 1 with `INVALID_INPUT` error: amount must be non-negative. (Note: clap may reject negative `f64` before the command handler is reached depending on argument parsing behavior — verify the error path either way.)
- [ ] `--amount 21000000000` (exceeds u64 RAO max) — expect exit code 1 with `INVALID_INPUT` error: "amount overflows u64 in RAO".
- [ ] `--amount abc` — clap parses `--amount` as `f64`; expect a parse error before the command handler is invoked.

#### Password prompt

- [ ] Verify password input is not echoed to the terminal (uses `rpassword`).
- [ ] Wrong password on an encrypted coldkey — expect exit code 1 with `CRYPTO_ERROR`: decryption failed or invalid password.

---

### Unit Test Verification

Run from the repo root:

```
cargo test
```

Expected: **49 tests pass**, 0 failures.

Verify the following test coverage is present and passing:

**TAO/RAO conversion (`commands::wallet_keys`)**
- [ ] `tao_to_rao_whole_number` — 1.0 TAO = 1,000,000,000 RAO
- [ ] `tao_to_rao_fractional` — 0.5 TAO = 500,000,000 RAO
- [ ] `tao_to_rao_zero` — 0.0 TAO = 0 RAO
- [ ] `tao_to_rao_large_amount` — 21,000,000 TAO = 21,000,000,000,000,000 RAO
- [ ] `tao_to_rao_negative_fails` — negative input returns error
- [ ] `tao_to_rao_nan_fails` — NaN input returns error
- [ ] `tao_to_rao_infinity_fails` — infinite input returns error
- [ ] `rao_to_tao_string_whole` — 1,000,000,000 RAO = `"1.0"`
- [ ] `rao_to_tao_string_fractional` — 1,500,000,000 RAO = `"1.5"`
- [ ] `rao_to_tao_string_zero` — 0 RAO = `"0.0"`
- [ ] `rao_to_tao_string_small` — 1 RAO = `"0.000000001"`
- [ ] `rao_to_tao_string_precise` — 100,500,000,000 RAO = `"100.5"`

**Coldkey decryption (`commands::wallet_keys`)**
- [ ] Encrypt/decrypt roundtrip — key encrypted with a password decrypts successfully with the same password
- [ ] Wrong password rejection — decryption with incorrect password returns a `CRYPTO_ERROR`

**Stake address resolution (`commands::stake`)**
- [ ] `resolve_address_ss58_validates` — valid SS58 passes and is returned unchanged
- [ ] `resolve_address_invalid_ss58_fails` — invalid address string returns error
- [ ] `resolve_address_both_provided_fails` — providing both `--wallet` and `--ss58` returns error
- [ ] `resolve_address_neither_provided_fails` — providing neither returns error

moorkh:
[cryptid] Barbarian review — stake-ops (PR #2)

I looked only at the stake-specific surface: `src/commands/stake.rs`, the stake branches in `src/cli.rs` and `src/main.rs`, and the deps added by commit `6194f5e`. wallet_keys is being reviewed under PR #3; I note cross-cutting interactions only.

Subtensor reference: https://github.com/opentensor/subtensor/blob/main/pallets/subtensor/src/macros/dispatches.rs and `pallets/subtensor/src/lib.rs` (Alpha / StakingHotkeys / NetUid definitions).

═══════════════════════════════════════════════════════════════
CRITICAL
═══════════════════════════════════════════════════════════════

## C-1  `list` decodes Alpha storage as `u64`, but it is `U64F64` shares → every stake is reported incorrectly (usually as 0)

`src/commands/stake.rs:131-143`

```rust
if let Ok(Some(val)) = storage.fetch(&alpha_query).await {
    if let Ok(dv) = val.to_value() {
        if let Some(amount) = dv.as_u128() {
            if amount > 0 {
                stakes.push(StakeEntry {
                    hotkey: hotkey_ss58.clone(),
                    netuid,
                    stake_tao: rao_to_tao_string(amount as u64),
                });
```

The doc comment at line 51 claims `SubtensorModule::Alpha` is a triple map `(hotkey, coldkey, netuid) -> u64`. It is not. From subtensor `pallets/subtensor/src/lib.rs:1427`:

```rust
pub type Alpha<T: Config> = StorageNMap<
    _,
    ( NMapKey<Blake2_128Concat, T::AccountId>,  // hot
      NMapKey<Blake2_128Concat, T::AccountId>,  // cold
      NMapKey<Identity,         NetUid>, ),     // subnet
    U64F64,                                     // Shares  <-- NOT u64
    ValueQuery,
>;
```

The stored value is `substrate_fixed::types::U64F64` — a 128-bit fixed-point number representing the coldkey's share of the hotkey/subnet stake pool. Two cascading defects:

1. **Denomination is wrong.** The value is shares, not alpha, not tao. Even if you read the raw 128 bits correctly, you still need `TotalHotkeyShares` / `TotalHotkeyAlpha` to convert shares → alpha → (via pool price) → TAO. None of that exists here.
2. **Decoding is wrong.** `dv.as_u128()` grabs the raw 128-bit integer of the fixed-point encoding. The code then does `amount as u64`, keeping only the lower 64 bits — i.e. the *fractional* part of the fixed-point number. The integer shares are silently discarded.

Proof — PoC illustrating the U64F64 raw layout and the effect of `as u64`:

```
5.0 shares raw=0x00000000000000050000000000000000
  btt interprets `amount as u64` = 0
  reported stake_tao = 0.0 TAO

5.5 shares raw=0x00000000000000058000000000000000
  btt interprets `amount as u64` = 9223372036854775808 (= 2^63)
  reported stake_tao = 9223372036 TAO

12345.0 shares raw=0x00000000000030390000000000000000
  btt interprets `amount as u64` = 0
  reported stake_tao = 0.0 TAO
```

For any whole-number share balance (the common case), `amount as u64 == 0`, and the `> 0` guard skips the entry entirely. The user sees an empty list and believes they hold nothing. For fractional shares, the reported TAO is a nonsense number in the 10^9–10^19 range.

Fix sketch: decode Alpha as `u128` (U64F64 raw), read `TotalHotkeyAlpha` and `TotalHotkeyShares`, compute `alpha = shares * total_alpha / total_shares`, then for a true TAO valuation multiply by the subnet's current alpha price from the swap pool. Until this is done, `btt stake list` must be removed or prefixed with a loud "untrusted / debug only" marker. Shipping it as-is guarantees user confusion and hides real balances.

## C-2  `remove` with `--all` reads Alpha shares as u64 and submits them as `amount_unstaked` → either "no stake found" or garbage-value unstake

`src/commands/stake.rs:327-391`

Same root cause as C-1. The `--all` path:

```rust
if let Ok(Some(val)) = storage.fetch(&alpha_query).await {
    if let Ok(dv) = val.to_value() {
        if let Some(amount) = dv.as_u128() {
            found_amount = amount as u64;
        }
    }
}
...
found_amount                         // u64, used directly
...
Value::u128(amount_rao as u128),     // passed as amount_unstaked
```

Outcomes for a user running `btt stake remove --wallet w --hotkey h --netuid 7 --all`:
- whole-number shares → `amount as u64 == 0` → code returns `invalid_amount("no stake found for this hotkey/netuid combination")`. User cannot unstake.
- fractional shares → `found_amount` is ~2^63 garbage → extrinsic submitted trying to unstake ≈9.2×10^18 alpha-rao → pallet rejects with NotEnoughStakeToWithdraw, or in a pathological case the pallet's saturating arithmetic clips and unstakes the full real balance. Either way, the amount the user asked for is not the amount on the wire, and the `tx_hash`/`amount_tao` in the success envelope is a lie.

The fallback path queries `SubtensorModule::Stake` which does not exist in current runtime (see H-2), so the fallback never helps.

Fix: same as C-1, plus — once you know the real alpha balance — pass it as `amount_unstaked` in the extrinsic.

═══════════════════════════════════════════════════════════════
HIGH
═══════════════════════════════════════════════════════════════

## H-1  `remove` denominates `--amount` as TAO but the pallet expects alpha → user specifies 1 TAO worth, btt sends 10^9 alpha-rao

`src/cli.rs:192-197` and `src/commands/stake.rs:383-390`

CLI help:
```rust
/// Amount in TAO (e.g. 10.5). Ignored if --all is set.
#[arg(long, required_unless_present = "all")]
amount: Option<f64>,
```

In `stake::remove`:
```rust
let tao = amount_tao.ok_or_else(...)?;
let rao = tao_to_rao(tao)?;               // TAO -> RAO (10^-9 TAO)
...
Value::u128(amount_rao as u128),          // passed as amount_unstaked
```

Subtensor signature (`pallets/subtensor/src/macros/dispatches.rs:760-766`):
```rust
pub fn remove_stake(
    origin: OriginFor<T>,
    hotkey: T::AccountId,
    netuid: NetUid,
    amount_unstaked: AlphaBalance,   // Alpha, 9 decimals, NOT TAO
) -> DispatchResult
```

btcli confirms the convention (`bittensor_cli/src/commands/stake/remove.py:812-819`): it builds an `Alpha(...)` balance object and passes `amount.rao` (alpha smallest unit) into `amount_unstaked`. btcli's user-facing flag is explicitly in alpha units and surfaces subnet price to translate TAO intent ↔ alpha amount.

btt's CLI advertises `--amount` as TAO and translates via `tao_to_rao`. On subnet 0 (root), alpha ≡ TAO 1:1 so this happens to be correct by accident. On every other subnet, 1 alpha ≠ 1 TAO, and typically differs by factors of 10×–1000× depending on the pool price. User types `--amount 5.0` meaning "5 TAO worth", code unstakes 5 alpha — on a subnet where alpha trades at 0.01 TAO this removes 0.05 TAO worth; on a subnet where alpha trades at 10 TAO this removes 50 TAO worth. Silent, money-material.

Proof — pallet signature + btcli reference + CLI help text mismatch, all quoted above.

Fix: two sane options.
- (a) Rename/document `--amount` as alpha units and require the user to know their subnet; expose a separate command to query the current price so the user can do the conversion.
- (b) Keep `--amount` in TAO as a convenience, but query the pool price at the head block and convert TAO → alpha before submission. Must also surface the computed alpha amount in the confirmation output so the user is not guessing.

Same issue applies to `add`, but is less severe there: `add_stake.amount_staked` really is `TaoBalance` (dispatches.rs:716-722), so `tao_to_rao` is correct for `add`. The slippage happens downstream in the pool. `add` is OK.

## H-2  `Stake` fallback storage does not exist on current subtensor → silent dead code that masks other failures

`src/commands/stake.rs:168-229` and `src/commands/stake.rs:355-373`

Both the list fallback and the `--all` fallback query `subxt::dynamic::storage("SubtensorModule", "Stake", ...)`. No such storage item exists in the current subtensor runtime — I grepped `pallets/subtensor/src/lib.rs` for `pub type Stake` and found only `TotalStake`, `AutoStakeDestination`, `StakeWeight`, `StakeThreshold`, `NominatorMinRequiredStake`, etc. No bare `Stake` double map. The `(hotkey, coldkey) -> u64` total-stake store was removed during the dTAO refactor; per-subnet alpha shares are the only ledger now.

subxt's dynamic `storage.fetch` for an unknown item will return an `Err`, which the code swallows with `if let Ok(Some(val)) = ...`. Result: the fallback always produces `None` / `0`, never helps, and hides any real decode error on the Alpha path because it silently produces "no result" instead of surfacing the underlying problem.

Proof — grep `pallets/subtensor/src/lib.rs` for `pub type Stake\b` returns nothing; only composite names like `TotalStake`, `StakeWeight`. The `Stake` double map is gone.

Fix: remove both fallback paths. They are not a safety net; they are a silencer.

## H-3  Hardcoded `for netuid in 0..64u16` misses every stake on subnets ≥ 64

`src/commands/stake.rs:120`

```rust
for netuid in 0..64u16 {
```

Mainnet has had more than 64 subnets for some time (`TotalNetworks` is a runtime StorageValue; current value is well over 64). A coldkey with a real position on netuid 73 sees it omitted from `btt stake list`, gets an empty list, and believes it has no stake. Same danger as C-1 — hidden balances.

Fix: query `SubtensorModule::TotalNetworks` once, iterate `0..total`. Or iterate `StakingHotkeys` → for each hotkey, iterate all netuids the subnet registry knows about. Better, switch to a runtime API call like `SubnetInfoRuntimeApi::get_stake_info_for_coldkey` (the doc comment at line 54 already imagines this but never implements it).

## H-4  Stake commands inherit the wallet_keys stdout password-prompt bug → JSON envelope corrupted

`src/commands/wallet_keys.rs:747-748` (called from `stake::add` and `stake::remove`)

`wallet_keys::read_password` uses `rpassword::prompt_password_stdout`, writing the prompt to stdout. Any `btt stake add …` invocation outputs `Enter coldkey password: ` on stdout, then (after input) the JSON success/error envelope. Any consumer piping `btt` through `jq` or into a script fails to parse. This is the same class as the H-3 finding already recorded on the PR #3 (wallet-keys) review.

Flagged here only as a cross-PR interaction: stake commands are the first non-wallet consumers of the decrypt path, so on stake-ops the bug is materially visible to users running staking flows. Fix belongs on the wallet-keys branch (`prompt_password_stderr`, or gate behind `atty::is(Stream::Stdout)`). Stake.rs needs no change once that lands — but it must land before stake-ops merges, or stake-ops will ship the regression.

═══════════════════════════════════════════════════════════════
MEDIUM
═══════════════════════════════════════════════════════════════

## M-1  `list` issues O(hotkeys × 64) sequential RPC calls with no per-call timeout

`src/commands/stake.rs:116-145`

The outer `at_latest()` is wrapped in `RPC_TIMEOUT` (30s), but each inner `storage.fetch(&alpha_query).await` inside the nested loop has no timeout. For a coldkey staking to, say, 30 hotkeys, this is 30 × 64 = 1920 sequential round-trips over a single WS connection. A hung RPC response hangs the entire command indefinitely. Even on a healthy endpoint this takes tens of seconds to minutes and is well over the `--quiet` UX threshold.

Fix: (a) wrap each fetch in `tokio::time::timeout`; (b) batch via a runtime API or multi-key fetch; (c) at minimum, parallelise with `futures::stream::iter(...).buffer_unordered(16)`.

## M-2  `extract_account_ids` / `extract_32_bytes` decode path has zero tests

`src/commands/stake.rs:448-486`

The decoded `Vec<AccountId32>` walk assumes `Value::at(idx)` descends into a composite sequence and that each element is itself a composite of 32 u8 primitives. This is plausible for subxt 0.38's dynamic output but is not verified anywhere. If the decoder yields a different shape — for example a `Primitive::Bytes` or a sequence of `Composite::Named(...)` — `extract_32_bytes` returns `None` for every element and `list` silently shows zero stakes. Given the project's `#![deny(clippy::unwrap_used)]` posture, this is also the place where failure will be swallowed rather than panicking.

Proof: the only stake tests are `resolve_address_*` (lines 510-535). No decoder fixture, no integration test against a fork.

Fix: add a unit test that constructs a scale_value::Value matching the `Vec<AccountId32>` shape produced by subxt and asserts `extract_account_ids` returns the expected bytes. Better, replace the manual walker with `subxt::utils::AccountId32` decoding via `val.as_type::<Vec<AccountId32>>()`.

## M-3  `--all` trusts the decrypted pair's public key, never cross-checks against `coldkeypub.txt`

`src/commands/stake.rs:320-322`

```rust
let pair = decrypt_coldkey_interactive(wallet)?;
let coldkey_ss58 = sp_core::crypto::AccountId32::from(PairTrait::public(&pair)).to_ss58check();
let coldkey_bytes = parse_ss58(&coldkey_ss58)?;
```

If the on-disk coldkey file has been tampered with or swapped, the decrypted pair's public key will not match `coldkeypub.txt`, but the code happily uses the decrypted pair's derived address to look up stakes and sign. Users trust `coldkeypub.txt` as the canonical public identity of the wallet; stake should verify the two agree before proceeding. Not an attack vector in the normal threat model, but a useful integrity check that costs two lines.

═══════════════════════════════════════════════════════════════
LOW
═══════════════════════════════════════════════════════════════

## L-1  `Stake` fallback fetches `StakingHotkeys` a second time

`src/commands/stake.rs:168-191`

`list()` fetches `StakingHotkeys` at line 95, then the fallback at line 173 fetches it again. One network round-trip wasted; the decoded hotkey list is already in scope and could be passed in. Dead after H-2 fix anyway.

## L-2  `stake_tao` field name is misleading regardless of C-1 fix

`src/commands/stake.rs:33`

Even if C-1 is fixed, the field will represent alpha (per-subnet token), not TAO. Post-dTAO, conflating the two is exactly the confusion H-1 warns about. Rename to `alpha` with a separate optional `tao_value` if the price conversion is done.

## L-3  SS58 round-trip in `remove`

`src/commands/stake.rs:321-322`

`AccountId32::from(public).to_ss58check()` then `parse_ss58(&coldkey_ss58)` — a pointless two-way conversion. Use the raw 32 bytes from `public().0` directly.

═══════════════════════════════════════════════════════════════
NIT
═══════════════════════════════════════════════════════════════

## N-1  Misleading doc comment on `list`

`src/commands/stake.rs:50-56` describes a `get_stake_info_for_coldkey` runtime-API fallback that does not exist in the code. Either implement it or delete the promise.

## N-2  Fallback error messages say "stake amount" in the remove path

`src/commands/stake.rs:375-379` — "no stake found" is the only failure signal for an empty `--all`. In the dTAO world, "no alpha found on this (hotkey, coldkey, netuid)" is more precise and less confusing than "no stake".

## N-3  Only 4 tests, all for `resolve_address`

Zero unit tests for the decoder, zero for the extrinsic-construction layer, zero for the TAO↔alpha intent, zero for error paths. The cross-cutting `tao_to_rao` tests live in wallet_keys and are fine, but stake.rs has no test asserting that a decoded Alpha value is transformed correctly, or that `--all` sends what was read. Given the defects above, this is the root cause of the whole review: nothing tested, nothing caught.

═══════════════════════════════════════════════════════════════
DEP DISCIPLINE
═══════════════════════════════════════════════════════════════

Stake commit `6194f5e` adds 7 new deps to `Cargo.toml` (bip39, rpassword, zeroize, xsalsa20poly1305, scrypt, rand, hex). All seven belong to wallet_keys, not to staking; stake.rs imports nothing from them directly. No finding for stake — but note that the stacked branch means the wallet-keys dep justification must land in PR #3's review. Stake itself adds no net new deps beyond subxt/sp-core/tokio/serde, which are already justified.

═══════════════════════════════════════════════════════════════
SUMMARY
═══════════════════════════════════════════════════════════════

| Sev | Count |
|-----|-------|
| CRITICAL | 2 |
| HIGH     | 4 |
| MEDIUM   | 3 |
| LOW      | 3 |
| NIT      | 3 |

VERDICT: **REQUEST_CHANGES**

The core staking ledger on current subtensor is per-subnet alpha *shares* in U64F64, bridged to TAO through a liquidity pool. This PR was written against a pre-dTAO mental model — `Alpha is u64`, `Stake is a double map`, `amount_unstaked is TAO` — none of which are true anymore. The code compiles because subxt's dynamic API will accept any shape the metadata permits, but the semantics are wrong end-to-end:

- `list` reports 0 TAO (or nonsense numbers) for real positions.
- `remove` with `--all` refuses to unstake real balances or submits garbage values.
- `remove --amount N` sends `N × 10^9` alpha units, not N TAO worth.
- stakes on netuids ≥ 64 are invisible.

Fixing the unit model (C-1, C-2, H-1) probably means rewriting the bodies of `list` and `remove`. I would pause, read the relevant sections of `pallets/subtensor/src/staking/*.rs` and btcli's `commands/stake/{add,remove,list}.py`, and re-draft against the current runtime. The existing CLI skeleton and address-resolution layer can be kept.

cryptid out.

moorkh:
[cryptid] clarification for the record: this PR's head commit `6194f5e` became reachable from `main` transitively via PR #3's wallet-keys branch (which was based on top of stake-ops). the content added by this PR — the pre-dTAO version of `src/commands/stake.rs` — was immediately replaced wholesale by PR #23 on 2026-04-13 via commit `c35cb0bf`, which rewrote the stake module against the dTAO runtime and closed all four CRITICAL+HIGH findings from this PR's round-1 barbarian review.

main as of now carries the PR #23 version of `stake.rs`, not the version from this PR. this PR's reachability-based merge state is an artifact of the shared-history between the stake-ops and wallet-keys branches during the early project scaffolding. the behavior is correct, the attribution is just a bit misleading.

PR #4, opened 2026-04-12
feat: resolve rpc endpoint from env vars with priority fallback

kompotkot:
## What

Integrate environment variable support for RPC endpoint configuration, enabling users to override connection endpoints without recompiling or passing CLI flags every time.

## Endpoint Resolution Priority

The `resolve_endpoint` function now enforces a strict 3-level priority chain:

1. **`--url` flag** *(highest)* — Explicit URL passed on the command line. Validated for `ws://` or `wss://` scheme and used immediately, overriding everything else.
2. **Environment variables** *(middle)* — If no `--url` is given, the matching env var is checked based on the selected network:
   - `RPC_DEFAULT_ENDPOINT` → used for `--network finney` or when no network is specified
   - `RPC_TEST_ENDPOINT`    → used for `--network test`
   - `RPC_LOCAL_ENDPOINT`   → used for `--network local`
3. **Hardcoded defaults** *(lowest)* — Fallback constants defined in `src/rpc.rs` are used when no flag or env var is set.

## How

- Added `dotenvy = "0.15"` — loads `.env` file automatically at startup (`main.rs`), file is optional and errors are silently ignored.
- Updated `resolve_endpoint()` in `src/rpc.rs` to call `std::env::var()` for each network branch before falling back to the constant.
- Added a new unit test `resolve_endpoint_uses_env_var_override` verifying env var override works correctly.
- Added `.env.sample` documenting the three supported variables.


moorkh:
# Review: PR #4 — env var support for RPC endpoints

Checked out, built, tested, clippy-clean. Then I fed it a few shapes of hostile input. Findings below.

---

## HIGH — env-var endpoints bypass `validate_url()` entirely

**File:** `src/rpc.rs:32-34, 39`

The `--url` flag path calls `validate_url(u)?` before returning. The env-var path does not. A value from `RPC_DEFAULT_ENDPOINT` / `RPC_TEST_ENDPOINT` / `RPC_LOCAL_ENDPOINT` is returned verbatim and handed straight to subxt.

This is the stated invariant of the module ("Returns an error for unrecognized network names **or invalid URL schemes**" — doc comment on `resolve_endpoint`). The env-var branch silently violates it.

**Proof** (on a fresh checkout of this PR, no `.env` file):

```
$ RPC_DEFAULT_ENDPOINT="ws://127.0.0.1:9944" cargo run -- chain info --network finney
{"ok":false,"error":{"code":"CONNECTION_FAILED",
 "message":"failed to connect to ws://127.0.0.1:9944: ... Connection refused"}}
```

The user asked for `finney`. They got silently redirected to loopback. subxt's own insecure-URL check lets `ws://` through for localhost, so btt's `wss://`-for-mainnet posture is gone the moment an env var or `.env` file is in play. An attacker who can plant a `.env` in the CWD (or drop one via any tool that writes to the working directory — editor templates, scaffolding scripts, a tarball extracted into `~/src/...`) can point a user who thinks they're querying mainnet Finney at an SSH-tunneled hostile node and forge arbitrary chain-state responses for `chain info` and `wallet balance`.

The dependency on `dotenvy` amplifies this: `dotenv()` auto-loads from CWD with zero prompting. The user does not have to opt in.

**Fix:** validate env var values the same way `--url` is validated, and apply network-specific hardening (no plain `ws://` for remote networks).

```rust
fn endpoint_from_env_or_default(var: &str, default: &str) -> Result<String, BttError> {
    match std::env::var(var) {
        Ok(v) if !v.is_empty() => {
            validate_url(&v)?;
            Ok(v)
        }
        _ => Ok(default.to_string()),
    }
}
```

And then in `resolve_endpoint`:

```rust
Some("finney") | None => endpoint_from_env_or_default("RPC_DEFAULT_ENDPOINT", DEFAULT_ENDPOINT),
Some("test")           => endpoint_from_env_or_default("RPC_TEST_ENDPOINT",    TEST_ENDPOINT),
Some("local")          => endpoint_from_env_or_default("RPC_LOCAL_ENDPOINT",   LOCAL_ENDPOINT),
```

Consider also gating `ws://` (plaintext) so it is only accepted for `--network local` or when host is `127.0.0.1`/`::1`/`localhost`. That is the invariant the hardcoded defaults already express; the env path should not quietly relax it.

---

## HIGH — empty env var is treated as a valid override, not a fallthrough

**File:** `src/rpc.rs:32-34, 39`

`std::env::var("X")` returns `Ok("")` when `X` is set but empty, and `Err(_)` only when it is unset. The current code uses `unwrap_or_else(|_| DEFAULT.to_string())`, which takes the empty string as a valid value.

**Proof:**

```
$ RPC_DEFAULT_ENDPOINT="" cargo run -- chain info --network finney
{"ok":false,"error":{"code":"CONNECTION_FAILED",
 "message":"failed to connect to : RPC error: relative URL without a base"}}
```

The user's intent with `FOO=""` is almost always "unset" — e.g., a shell conditional, a Docker compose file stripping a var, a CI matrix. Current behavior surprises the caller.

**Fix:** the same helper above (`Ok(v) if !v.is_empty() => ...`) handles both this and finding HIGH-1 in one stroke.

---

## MEDIUM — test pollution: three tests read `RPC_DEFAULT_ENDPOINT`, only one sets/clears it, none synchronize

**File:** `src/rpc.rs:92-120, 160-166`

After this change, these three tests all inspect the same global:

- `resolve_endpoint_finney_returns_finney_url` — asserts `== DEFAULT_ENDPOINT`
- `resolve_endpoint_none_defaults_to_finney`  — asserts `== DEFAULT_ENDPOINT`
- `resolve_endpoint_uses_env_var_override`    — sets `RPC_DEFAULT_ENDPOINT` to a custom value, then removes

`cargo test` runs tests in parallel by default. If the override test's `set_var` happens after one of the baseline tests reads the function but before it asserts — or if `set_var` hasn't been cleaned up yet when a baseline test runs — the baseline tests flip to failing. I couldn't force a flake in 20x8-thread runs here, but it's a latent race and your own in-code comment ("Mutating env vars in tests can race if run in parallel") acknowledges it. That comment is not a fix, it is a warning label on an unfenced edge.

There is also a second-order hazard: on Rust 2024 (which this crate will migrate to eventually), `std::env::set_var` is `unsafe` precisely because of this class of races. The current tests will not compile cleanly after that edition bump without an `unsafe {}` block and the reasoning made explicit.

**Fix:** serialize env-touching tests. Three options, in order of preference:

1. Force single-threaded for env tests via a mutex in the test module:
   ```rust
   use std::sync::Mutex;
   static ENV_LOCK: Mutex<()> = Mutex::new(());
   // acquire ENV_LOCK at the top of any test that reads/writes RPC_*_ENDPOINT
   ```
   and acquire it in all three affected tests, not just the new one. Use `.lock().unwrap_or_else(|e| e.into_inner())` so a panic in one test doesn't poison the next.

2. Inject env access via a small trait / function parameter so tests don't touch real env at all. Overkill for this codebase but cleanest.

3. Run the env tests behind `#[ignore]` with a dedicated `cargo test -- --ignored --test-threads=1` CI invocation. Least invasive, ugliest ergonomics.

Option 1 is the right call here.

---

## MEDIUM — `dotenvy` is not justified; a `.env` in CWD is an implicit trust boundary

**File:** `Cargo.toml:15`, `src/main.rs:16`

The PR description frames `dotenvy` as "loads `.env` file automatically at startup". The goal stated in the priority chain ("env vars override defaults") is fully satisfied by `std::env::var` alone — the user exports `RPC_DEFAULT_ENDPOINT=...` in their shell or CI config and it works. `dotenvy` exists only to auto-load a file from `$CWD`.

Two problems with that:

1. **Security.** `$CWD` is not a trusted directory. A user running `btt` from inside a project tree they cloned from GitHub gets whatever endpoint the repo's `.env` says. Combined with the missing validation above, this is a supply-chain redirect primitive. If you want file-based config, the right shape is an **explicit path** (e.g., `~/.config/btt/config.toml`, 0600, parsed intentionally), not "whatever happens to be in PWD".

2. **Dependency justification.** `dotenvy` itself has zero transitive deps (verified via `cargo tree -p dotenvy`), so the bloat cost is small. But "small dep" is not an argument for a dep whose sole feature is the thing you didn't actually need.

**Fix:** drop `dotenvy` entirely. Read env vars with `std::env::var` (which the code already does via `unwrap_or_else`). If a `.env`-style workflow is desired for developers, document in README: `set -a; source .env; set +a; cargo run -- ...`. That keeps the "which directory am I trusting" decision in the user's hands, not in the binary's.

If you keep `dotenvy`, at minimum:
- Only load from an explicit path (`dotenvy::from_path(...)`) pointed at `~/.config/btt/.env` or similar.
- Do not load from CWD.
- Log (to stderr) when a file is loaded, so the user can see "I just picked up `.env` from /tmp/sketchy-dir".

---

## LOW — `.gitignore` adds `Cargo.lock` for a binary crate

**File:** `.gitignore:12`

The toptal-generated template that was pasted in ignores `Cargo.lock`. `btt` is a binary (`src/main.rs`, `[[bin]]`-shaped crate). Cargo's own guidance: commit `Cargo.lock` for binaries, don't for libraries. `Cargo.lock` is currently tracked (`git ls-files Cargo.lock` confirms), so the ignore line is inert today — but it is a trap. If anyone later does `git rm --cached Cargo.lock` for any reason (cleanup script, reset, accidental `git clean -fdx` after untracking), the file stops being tracked and nobody notices until a reproducible-build bug bites.

**Fix:** remove the `Cargo.lock` line from `.gitignore`. Or remove the whole toptal preamble — it imports a bunch of VSCode rules and Rust conventions you didn't write and may not agree with — and hand-write the five lines this project actually needs.

---

## LOW — `.gitignore` pattern `*.env` is broader than `.env`

**File:** `.gitignore:45`

`*.env` ignores `.env`, `prod.env`, `dev.env`, `anything.env`. That may be intentional, but if it is, say so in a comment. More importantly, `.env.sample` is **not** matched (the `*` in `*.env` requires at least one character before the dot, and `.env.sample` doesn't end in `.env`), so the sample file stays tracked as intended. Verified with `git check-ignore -v`.

**Fix:** if the intent is only `.env`, write `.env`. If the intent is the broader glob, a one-line comment would save the next reviewer the 30 seconds I just spent.

---

## NIT — trailing whitespace and missing final newlines

- `src/main.rs:17` — line is `"    "` (four spaces) after `dotenvy::dotenv().ok();`. Visible in the diff. `cargo fmt` will strip it.
- `.env.sample` — no trailing newline (`\ No newline at end of file` in the diff).
- `.gitignore` — no trailing newline.

**Fix:** `cargo fmt`, and add `\n` to the two new files.

---

## NIT — doc comment on `resolve_endpoint` is now inaccurate

**File:** `src/rpc.rs:24-25`

```
/// Resolve a URL from explicit --url or --network shorthand.
```

It now also resolves from env vars. Update the doc.

**Fix:**

```rust
/// Resolve a URL from (in priority order): explicit --url, env var
/// (RPC_{DEFAULT,TEST,LOCAL}_ENDPOINT), or the hardcoded default for the
/// selected network. Returns an error for unrecognized network names or
/// invalid URL schemes (including schemes supplied via env var).
```

---

## What works

- Build is clean, 17 tests pass, clippy clean on `--all-targets`.
- `dotenvy` has zero transitive deps — bloat impact is minimal.
- The priority chain (flag > env > default) matches the PR description in the happy path.
- `--url` still calls `validate_url()`, so the explicit-flag path is unchanged and correct.
- `.env.sample` is committed and not accidentally ignored.

---

## Verdict

Two HIGH findings (unvalidated env endpoints + empty-string-as-value) that together turn a `.env` in the working directory into a silent endpoint redirect. These are blocking.

**VERDICT: REQUEST_CHANGES**

moorkh:
## Severity upgrade: dotenvy finding escalated to HIGH

Per project policy (btt exists to minimize supply chain attack surface), any unnecessary third-party dependency is a HIGH severity finding, not MEDIUM.

Reclassifying **M-2 (dotenvy is unjustified)** → **H-3 (unnecessary third-party dependency)**.

The priority chain goal (--url > env vars > hardcoded defaults) is fully satisfied by `std::env::var()` from the standard library. `dotenvy` exists only to auto-load `.env` from the current working directory, which is exactly the attack surface exploited by H-1 (env var endpoints bypassing validation). The dependency is both unnecessary and a direct enabler of the higher-severity finding.

Updated verdict: **REQUEST_CHANGES** with three HIGH findings (H-1, H-2, H-3). All three must be addressed before merge.

kompotkot:
I decided to revert the env vars support based on the code review findings. The main reason is that the env var approach bypasses `validate_url()` entirely, which poses a security risk because it allows plain `ws://` schemes on remote networks. This acts as a silent endpoint redirect.\n\nAdditionally, auto-loading from `.env` in the CWD introduces a trust boundary issue. For now, we will stick to the explicit `--url` flag, which correctly enforces validation and network invariants.

moorkh:
[cryptid] re-reviewed after revert (ffaaed1).

the diff is now 37 additions / 1 deletion, all in `.gitignore`. the env-var trust boundary problem is gone — `dotenvy` dep dropped, `.env` CWD auto-load dropped, `validate_url()` bypass no longer reachable. H-1, H-2, H-3 from the prior review are all moot. good call on reverting rather than papering over.

remaining notes on the gitignore itself (none blocking):

- `target/` now matches `target/` in any subtree, not just root. for btt this is fine; the toptal template uses the non-anchored form and it catches stray targets in git-subtree workspaces.
- `*.env` is broader than `.env` — matches `prod.env`, `dev.env`, etc. reading your earlier slack message, that's intentional, so leaving as-is. a one-line comment like `# any *.env file, not just .env` would save a future reviewer 30 seconds but this is a nit.
- `Cargo.lock` is *not* in the ignore list (good — it's a binary crate, lockfile should be tracked). verified.
- no secrets, no credentials, no network endpoints, no trust boundaries introduced.
- `.env.sample` still committed (`*.env` glob does not match it — the pattern needs at least one character before `.env`, and `.env.sample` doesn't end in `.env`). verified with `git check-ignore`.

**verdict: APPROVE**. merging.

Issue #6, opened 2026-04-13
cryptid/dep-audit: dependency checksum tripwire backed by ken


moorkh:
[cryptid]

## Goal

Detect a class of supply-chain attack against btt's dependencies that `cargo audit` and `cargo deny` do not catch on their own: **same dep, same version, different bytes**. If an adversary republishes a crate at a version we already trust (via a registry that allows overwrites, a yank+republish flow, a compromised maintainer account on a private index, a git-source dep whose ref moved, etc.), the version pin in `Cargo.lock` is unchanged and standard auditing tools see nothing wrong. We want a tripwire that fires when the *bytes* of a dep at a known version diverge from the bytes we previously observed.

This is a defense-in-depth measure on top of the existing planned dep audit work (`cargo audit` + `cargo deny` + `cargo outdated`, surfaced via PR-comment workflow). It addresses the failure mode that destroyed btcli in 2024 — a malicious update that rode on a maintainer-trust pathway rather than a discoverable CVE.

## Approach

Maintain a database keyed by `(dep_name, version)` mapping to a sha256 of the dep's source tarball / git tree / vendored content (the exact content that ends up in our build). On every CI run:

1. Parse `Cargo.lock` for the full transitive dep set + sources
2. For each `(name, version)` pair, compute the current canonical sha256 from the registry / git ref
3. Look up the entry in ken
4. If the entry exists and the checksum matches → silent pass
5. If the entry exists and the checksum *differs* → loud alarm. **especially** if the version is unchanged. this is the tripwire signal.
6. If the entry does not exist (new dep / new version) → record the current checksum, surface the dep in the PR audit comment as a new entry to be reviewed under `btt-dep-discipline`

The DB lives in ken (https://github.com/zomglings/ken), under a publication kind we'll define (probably `btt-dep-checksum`). Each publication's key is `<dep>@<version>`, value is the checksum + first-seen timestamp + first-seen-in-PR.

## Requirements (will be appended as we discuss)

To be filled in as the design conversation continues. See the comments on this issue for the running list.

### Initial requirements (round 1)

- [ ] Use ken as the persistent store for `(dep, version) → checksum` entries.
- [ ] Detect checksum disagreement on a known `(dep, version)` pair as a CRITICAL finding. No version change is the worst case.
- [ ] First-seen entries: record the checksum, do not block, but surface them in the dep audit PR comment for cryptid review.
[...]

PR #10, opened 2026-04-13
ci(dep-audit): basic dep audit workflow (cargo-audit + cargo-deny + cargo-outdated)

moorkh:
[cryptid]

Adds the basic dependency audit workflow for btt. Runs `cargo-audit`, `cargo-deny`, and `cargo-outdated` against the resolved dep tree on every PR, every push to main, weekly, and on workflow_dispatch. Emits a Markdown summary as a PR comment.

Per AGENTS.md principle 1, every dep is an attack surface until proven essential. This workflow is the first line of defense; the dep-checksum tripwire (issue #6) is the second, layered on later as a separate workflow.

## Files

- `.github/workflows/dep-audit.yml` — the workflow definition. Triggers on `pull_request`, `push: [main]`, weekly schedule, and `workflow_dispatch` (with a `verbose` input). Uses `actions/checkout@v4`, `actions/cache@v4`, `actions/upload-artifact@v4`, `actions/github-script@v7`, `dtolnay/rust-toolchain@stable`. All first-party except `dtolnay/rust-toolchain@stable` which is the de-facto standard for rust-on-actions.
- `scripts/dep-audit/audit.sh` — main script. Runs the three tools, formats their output as Markdown, exits non-zero on any cargo-audit advisory or cargo-deny error. cargo-outdated is informational only. Locates the workspace root from its own path so it can be run from any directory.
- `scripts/dep-audit/deny.toml` — cargo-deny configuration. License allowlist (MIT, Apache-2.0, BSD variants, ISC, Unicode-3.0, Zlib, CC0-1.0, MPL-2.0, 0BSD), `unknown-registry = deny`, `unknown-git = deny`, `yanked = deny`, `wildcards = deny`. Permissive at first; tightens as we discover what the project will and will not allow.
- `scripts/dep-audit/README.md` — documentation.

## How the PR comment posting works

The workflow uses `actions/github-script@v7` (first-party Github action) to find a previous `cryptid-dep-audit` comment by an HTML marker (`<!-- cryptid-dep-audit -->`) and update it in place if found, or create a new one if not. This avoids cluttering the PR with a new comment on every push.

The comment body is prefixed `[cryptid] dep-audit run <run-id>` per the identity rule (every cryptid communication on every surface starts with `[cryptid]`).

## What this PR does NOT include

- The dep-checksum tripwire (issue #6) — different scope, lands as a separate workflow after this one.
- Any change to `Cargo.toml` or `Cargo.lock` — this is purely additive CI infrastructure.
- Any change to btt's binary behavior — runs only in CI.

## Test plan

- [x] `bash -n scripts/dep-audit/audit.sh` — shell syntax clean
- [x] `python3 -c "import yaml; yaml.safe_load(open('.github/workflows/dep-audit.yml'))"` — yaml syntax clean
- [ ] First CI run on this PR exercises the workflow end-to-end. Expected: PASS (current dep tree has no known advisories, license allowlist covers all current deps, no banned crates)
- [ ] Manual local run after `cargo install --locked cargo-audit cargo-deny cargo-outdated` reproduces the same report
- [ ] Comment posting works: a comment appears on this PR after the first workflow run, and updates in place on subsequent pushes

## Merge

Plain merge commit per AGENTS.md principle 7. Squash button untouched.

## Related

- AGENTS.md principle 1 — dependency discipline
- Issue #6 — dep-checksum tripwire (the next layer; this PR is the prerequisite)
- The btcli July 2024 supply chain attack — the failure mode this defends against

### DIFF
--- a/.github/workflows/dep-audit.yml
+++ b/.github/workflows/dep-audit.yml
@@ -65,14 +65,22 @@ jobs:
           VERBOSE: ${{ inputs.verbose && '1' || '0' }}
         run: |
           chmod +x scripts/dep-audit/audit.sh
+          # The script's exit code is captured into the report's STATUS line
+          # and re-checked by the final step. Bash -e (the default for
+          # GH Actions run blocks) would abort here on script failure and
+          # skip every subsequent step in this same block, so we suspend
+          # it around the script invocation only.
+          set +e
           scripts/dep-audit/audit.sh > audit-report.md 2> audit-stderr.log
-          # Always print the report to the job log
+          script_exit=$?
+          set -e
           echo '--- audit report ---'
           cat audit-report.md
           if [ "${VERBOSE}" = "1" ]; then
             echo '--- stderr ---'
             cat audit-stderr.log
           fi
+          echo "audit_exit=${script_exit}" >> "$GITHUB_OUTPUT"
 
       - name: Find existing comment
         id: find-comment
@@ -126,11 +134,22 @@ jobs:
           retention-days: 90
 
       - name: Fail job on critical findings
+        if: always()
         run: |
           # The audit script exits non-zero if any cargo-audit advisory
           # is present, or if cargo-deny errors out. Re-check the report
-          # for an explicit FAIL marker.
+          # for an explicit STATUS line, and treat a missing report as a
+          # failure too (the script crashed before writing it).
+          if [ ! -s audit-report.md ]; then
+            echo 'audit-report.md missing or empty; the audit script did not run to completion'
+            exit 1
+          fi
+          if ! grep -q '^STATUS: ' audit-report.md; then
+            echo 'audit-report.md has no STATUS line; the audit script crashed mid-run'
+            exit 1
+          fi
           if grep -q '^STATUS: FAIL' audit-report.md; then
             echo 'dep audit reported FAIL'
             exit 1
           fi
+          echo 'dep audit reported PASS'
--- a/.github/workflows/dep-audit.yml
+++ b/.github/workflows/dep-audit.yml
@@ -1,0 +1,136 @@
+name: dep-audit
+
+# Runs cargo-audit, cargo-deny, and cargo-outdated against the repo's
+# dependency tree, and posts a structured summary as a PR comment. This
+# is the basic dep-audit pass; the dep-checksum tripwire (issue #6) is
+# layered on later as a separate workflow.
+#
+# Per AGENTS.md principle 1 (dependency discipline), every dep is an
+# attack surface until proven essential. This workflow surfaces the
+# information; the human reviewer enforces the policy.
+
+on:
+  pull_request:
+  push:
+    branches: [main]
+  schedule:
+    # Sundays at 04:00 UTC. Catches advisories that land between PRs.
+    - cron: '0 4 * * 0'
+  workflow_dispatch:
+    inputs:
+      verbose:
+        description: 'Verbose output (dump full cargo tree and raw tool output)'
+        type: boolean
+        default: false
+
+permissions:
+  contents: read
+  pull-requests: write
+
+jobs:
+  audit:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Install Rust toolchain
+        uses: dtolnay/rust-toolchain@stable
+
+      - name: Cache cargo audit / deny / outdated binaries
+        id: cargo-tools-cache
+        uses: actions/cache@v4
+        with:
+          path: |
+            ~/.cargo/bin/cargo-audit
+            ~/.cargo/bin/cargo-deny
+            ~/.cargo/bin/cargo-outdated
+          key: cargo-audit-tools-v1
+
+      - name: Install cargo-audit
+        if: steps.cargo-tools-cache.outputs.cache-hit != 'true'
+        run: cargo install --locked cargo-audit
+
+      - name: Install cargo-deny
+        if: steps.cargo-tools-cache.outputs.cache-hit != 'true'
+        run: cargo install --locked cargo-deny
+
+      - name: Install cargo-outdated
+        if: steps.cargo-tools-cache.outputs.cache-hit != 'true'
+        run: cargo install --locked cargo-outdated
+
+      - name: Run dep audit
+        id: audit
+        env:
+          VERBOSE: ${{ inputs.verbose && '1' || '0' }}
+        run: |
+          chmod +x scripts/dep-audit/audit.sh
+          scripts/dep-audit/audit.sh > audit-report.md 2> audit-stderr.log
+          # Always print the report to the job log
+          echo '--- audit report ---'
+          cat audit-report.md
+          if [ "${VERBOSE}" = "1" ]; then
+            echo '--- stderr ---'
+            cat audit-stderr.log
+          fi
+
+      - name: Find existing comment
+        id: find-comment
+        if: github.event_name == 'pull_request'
+        uses: actions/github-script@v7
+        with:
+          result-encoding: string
+          script: |
+            const { data: comments } = await github.rest.issues.listComments({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              issue_number: context.issue.number,
+            });
+            const marker = '<!-- cryptid-dep-audit -->';
+            const existing = comments.find(c => c.body && c.body.includes(marker));
+            return existing ? String(existing.id) : '';
+
+      - name: Post or update PR comment
+        if: github.event_name == 'pull_request'
+        uses: actions/github-script@v7
+        with:
+          script: |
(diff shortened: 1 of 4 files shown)


moorkh:
[cryptid] barbaric review of PR #10 — `ci(dep-audit): basic dep audit workflow`

This PR shipped broken. The CI run it triggered — run [24346715282](https://github.com/ErgodicLabs/btt/actions/runs/24346715282) — failed on first contact. The failure is not a policy failure (real advisory found, good). It is a configuration failure (wrong `cargo deny` invocation). Do not merge.

---

## CRITICAL

### 1. Wrong `cargo deny` flag position — workflow fails on every run

**File:** `scripts/dep-audit/audit.sh`, line 82

```bash
cargo deny --config "${DENY_CONFIG}" check 2>&1
```

This is wrong. `cargo deny` does not accept `--config` as a top-level flag before the subcommand. The correct form is:

```bash
cargo deny check --config "${DENY_CONFIG}" 2>&1
```

**Proof — from the actual CI artifact (`audit-report.md`, line 15912):**

```
error: unexpected argument '--config' found

  tip: 'check --config' exists

Usage: cargo-deny [OPTIONS] <COMMAND>
```

`cargo-deny` ran, rejected `--config` at the wrong position, exited 1. The deny check did not execute. The script reported `FAIL` for the wrong reason. Every dep-deny policy (license allowlist, source allowlist, ban rules) was silently skipped.

---

### 2. `direct_count` and `total_count` are always 0 — grep patterns broken

**File:** `scripts/dep-audit/audit.sh`, lines 47–49

```bash
direct_count="$(cargo tree --depth 1 --quiet 2>/dev/null | grep -c '^[└├]──' || echo 0)"
total_count="$(cargo tree --quiet 2>/dev/null | grep -c '^[│ ]*[├└]──' || echo 0)"
emit "Total: ${direct_count} direct, ${total_count} entries in resolved tree (transitive)."
```

`cargo tree` emits ANSI dim-escape sequences around its box-drawing characters: `[2m├──[0m`, not bare `├──`. The grep patterns match literal `├──` and `└──`, which never appear. Both counts are always 0.

**Proof — from the actual CI artifact (`audit-report.md`, lines 22–24):**

```
Total: 0
0 direct, 0
0 entries in resolved tree (transitive).
```

The tree above it clearly shows 12 direct deps. The count is wrong.

---

## HIGH

### 3. `deny.toml` license clarify for `ring` — `OpenSSL` not in allowlist

**File:** `scripts/dep-audit/deny.toml`, lines 43–45 and 25–38

```toml
[[licenses.clarify]]
name = "ring"
expression = "MIT AND ISC AND OpenSSL"
license-files = [{ path = "LICENSE", hash = 0xbd0eed23 }]
```

The clarify entry declares `ring`'s license as `MIT AND ISC AND OpenSSL`. The `OpenSSL` SPDX identifier is not in the `[licenses] allow` list. When `cargo deny check licenses` actually runs (after the `--config` bug is fixed), it will fail on `ring`. `ring` is a transitive dep of `sp-core` and is present in `Cargo.lock`.

The hash `0xbd0eed23` is also suspect — it appears to be copy-pasted from the cargo-deny documentation template rather than computed against the actual `ring` 0.17.14 LICENSE file. If it's wrong, cargo-deny will reject the clarification entirely.

Fix: either add `"OpenSSL"` to the allowlist, or change the expression to `"MIT AND ISC"` if the OpenSSL portions are actually covered under one of the other identifiers, and recompute the hash.

### 4. `cargo audit` output contains raw ANSI escape sequences in the Markdown report

**File:** `scripts/dep-audit/audit.sh`, line 58–60

```bash
audit_output="$(cargo audit --quiet 2>&1)"
emit "${audit_output}"
```

`cargo audit` writes colored terminal output. `--quiet` suppresses progress but not ANSI codes. The captured output is embedded verbatim inside a fenced code block. The CI artifact confirms this — the posted report contains `[0m[1m[31mCrate:` rather than `Crate:`. The GitHub PR comment will be visually polluted with escape sequences.

Fix: pipe through `cargo audit --quiet 2>&1 | sed 's/\x1B\[[0-9;]*m//g'`, or pass `--color never` if available, or set `NO_COLOR=1`.

### 5. `Fail job on critical findings` step is unreachable dead code

**File:** `.github/workflows/dep-audit.yml`, lines 128–136

The step has no `if:` condition. It defaults to `if: success()`. The `Run dep audit` step (line 62) exits 1 when findings are present. A failed step causes all subsequent steps without `if: always()` to be skipped.

**Proof — from the CI run log:**

```
X Run dep audit
- Find existing comment        (skipped)
- Post or update PR comment    (skipped)
✓ Upload audit artifact        (if: always())
- Fail job on critical findings (skipped)
```

The step never executes. Its comment says "Re-check the report for an explicit FAIL marker" — a backstop that never fires. The job does fail (because the script exits 1), so the practical effect is benign, but the step also fails to catch the false-negative case described in the prompt: if the script crashes before writing `STATUS: FAIL`, the exit code still propagates, but the intent of the step (verify the marker exists) is never tested.

Either add `if: always()` and make the step explicitly check both that the file exists and that it contains a STATUS line, or delete the step entirely since the script's exit code already propagates.

---

## MEDIUM

### 6. Cache key has no OS or rustc version — stale binaries across runner upgrades

**File:** `.github/workflows/dep-audit.yml`, line 48

```yaml
key: cargo-audit-tools-v1
```

No runner OS, no rustc version. Compiled `cargo-audit`/`cargo-deny`/`cargo-outdated` binaries from one rustc version will be served to jobs running a different rustc. When `ubuntu-latest` bumps or the stable toolchain advances, the cache will serve ELF binaries compiled against a mismatched glibc or rustc ABI until someone manually invalidates by bumping `v1`.

Fix: include `runner.os` and the rustc version in the cache key:

```yaml
key: ${{ runner.os }}-cargo-audit-tools-v1-${{ steps.toolchain.outputs.rustc-hash }}
```

### 7. Fork PRs will fail noisily on the comment step

**File:** `.github/workflows/dep-audit.yml`, lines 93–116

For PRs opened from forks, `GITHUB_TOKEN` is restricted to read-only even with `pull-requests: write` declared. The `issues.createComment` call will 403. The job will fail for an infrastructure reason unrelated to the dep audit itself, and the contributor gets a confusing error.

Fix: add `if: github.event.pull_request.head.repo.full_name == github.repository` to the post-comment step, or use the `pull_request_target` trigger with appropriate caution.

---

## LOW

### 8. `dtolnay/rust-toolchain@stable` not SHA-pinned

**File:** `.github/workflows/dep-audit.yml`, line 38

`actions/checkout`, `actions/cache`, etc. are first-party and SHA-pinned behind the version tag by GitHub's infrastructure. `dtolnay/rust-toolchain` is a community action. Using `@stable` means a compromised or malicious push to that repo's `stable` branch installs arbitrary code on the runner with access to `GITHUB_TOKEN`.

This is acceptable risk for a read/comment-only workflow, but it should be documented as a deliberate decision. Better: pin to a specific commit SHA.

### 9. `git rev-parse --abbrev-ref HEAD` emits `HEAD` in CI (detached HEAD)

**File:** `scripts/dep-audit/audit.sh`, line 33

In CI, checkout leaves the repo in detached HEAD state. The report header reads `Run against 144a1a4 on HEAD.` — not the branch name. Minor cosmetic defect, but misleading. Use `$GITHUB_REF_NAME` in CI or accept it.

### 10. Node.js 20 deprecation warning on all action pins

The CI run produced deprecation warnings for `actions/checkout@v4`, `actions/cache@v4`, `actions/upload-artifact@v4`. These actions run on Node.js 20 which is being removed from runners on 2026-09-16. Not a blocker today, but will become one before the year is out.

---

## Summary table

| # | Severity | Finding |
|---|----------|---------|
| 1 | CRITICAL | `cargo deny --config` wrong argument position — deny check never runs |
| 2 | CRITICAL | Dep count grep patterns broken by ANSI codes — always reports 0 |
| 3 | HIGH | `ring` clarify uses `OpenSSL` SPDX id not in allowlist; hash unverified |
| 4 | HIGH | ANSI escape sequences embedded in Markdown PR comment output |
| 5 | HIGH | `Fail job on critical findings` step is skipped by default `if: success()` |
| 6 | MEDIUM | Cache key missing OS and rustc version — stale binary risk |
| 7 | MEDIUM | Fork PR comment step will 403 and fail the job |
| 8 | LOW | `dtolnay/rust-toolchain@stable` not SHA-pinned |
| 9 | LOW | Detached HEAD in CI produces `HEAD` as branch name in report |
| 10 | LOW | Node.js 20 deprecation warnings on action pins |

---

## Verdict: REQUEST CHANGES

The workflow failed on its first CI run for a non-policy reason (wrong argument order to `cargo deny`). The dep count reporting is broken. The `ring` license clarification will produce a second failure once the `--config` position is fixed. Three critical/high issues would cause the CI to either fail for wrong reasons or silently skip policy checks. Do not merge until findings 1–5 are addressed and the workflow passes end-to-end on the current dep tree without configuration errors.

PR #12, opened 2026-04-13
feat(wallet): --password-file + btcli-compat CI workflow

moorkh:
## Summary

Two tightly coupled changes that together turn btt's btcli keyfile
compatibility from a "should work" claim into a CI-enforced invariant.

1. **`--password-file <PATH>`** on `btt wallet create`, `new-coldkey`,
   `regen-coldkey`, and `sign`. Reads the coldkey password from a
   mode-0600 regular file instead of prompting. Full unit tests for
   the file reader: happy path, LF/CRLF stripping, past-newline
   discard, world/group-readable refusal, missing file, FIFO
   rejection, directory rejection, tilde expansion.

2. **`scripts/btcli-compat/check.py` + `.github/workflows/btcli-compat.yml`**.
   A standalone python verifier that, for each of six deterministic
   `(password, plaintext)` vectors, proves both directions of
   envelope compatibility:

   - btt encrypts (`wallet regen-coldkey --password-file`), pynacl
     decrypts, mnemonic recovered.
   - pynacl encrypts (a repack of a bootstrap btt blob), btt decrypts
     via `wallet sign --password-file`, signature is well-formed.

   The verifier imports only `nacl.secret`, `nacl.utils`,
   `nacl.exceptions`, and `argon2.low_level` — libsodium and reference
   argon2 wrappers. **It never imports `bittensor.*` or
   `btwallet.*`.** Pins are `--require-hashes` with sha256 values
   verified against the pypi JSON API on 2026-04-13.

This PR also restores `cargo build` on `main`. The wallet-keys merge
left stale `print_success(&result, pretty, quiet)` three-arg calls in
main.rs after an earlier patch dropped the third argument; main has
been failing to compile since that merge. The password-file plumbing
rewrites every affected match arm, so the fix lands as a side effect.

## Why this exists

btt's coldkey on-disk format has to stay byte-for-byte wire-compatible
with btcli/btwallet, because users need to be able to migrate wallets
without re-encrypting anything. But btt also cannot import or run
`bittensor.*` at any point in its lifecycle, because that's the
supply chain (bittensor 6.12.2, July 2024) this project exists to
route around.

Those two constraints are in tension. The compat workflow holds
both by verifying envelope compatibility against *reference crypto
primitives* (pynacl, argon2-cffi — thin wrappers over libsodium and
the reference argon2 C), never against bittensor. And by keeping the
verifier in CI only, never in dev environments, never in release
artifacts.

## Files

- `src/commands/password_file.rs` — new, 320 lines (130 code + 190
  tests)
- `src/commands/mod.rs` — one-line module registration
- `src/cli.rs` — `--password-file` on 4 subcommands with the full
  security warning in each help text
- `src/main.rs` — `resolve_coldkey_password` helper + rewrites the
  four match arms; drops the stale quiet arg from `print_success`
  calls
- `src/commands/skill.rs` — documents `--password-file` for every
  command that accepts it, adds a `#### --password-file` section
- `README.md` — new "Non-interactive automation" section and a
  "btcli format compatibility" section pointing at the workflow
- `AGENTS.md` — single line under principle 4 noting the
  CI-only verification mechanism
- `scripts/btcli-compat/check.py` — new, 654 lines. Deterministic
  vector generation, bidirectional envelope test, JSON report, raw
  blob artifacts
- `scripts/btcli-compat/btcli-pins.txt` — new, 57 lines. Pinned
  sha256 hashes for pynacl, argon2-cffi, argon2-cffi-bindings,
  cffi, pycparser. Verified against the pypi JSON API
- `scripts/btcli-compat/README.md` — new, ~90 lines. Design, threat
  model, local-run instructions, "when this check fails" runbook
- `.github/workflows/btcli-compat.yml` — new, 119 lines. PR paths
  filter, push to main, Sunday cron, `workflow_dispatch` with
  `vector_count`/`verbose` inputs. Uploads raw blobs on success and
  failure for 90 days

## Local verification performed

- `cargo build` — clean (was 17 errors on main before this PR)
- `cargo build --release` — clean
- `cargo test --release` — 67 passed, 1 ignored (pre-existing), 0
  failed. All 13 new `password_file` tests pass, including the unix-
  only FIFO rejection test.
- `cargo clippy --all-targets -- -D warnings` — clean
- `python3 -m py_compile scripts/btcli-compat/check.py` — clean
- **End-to-end compat run** against the built btt binary with
  pynacl 1.5.0 + argon2-cffi installed on this machine:

  ```
  seed: 8eb5c39e81b64bcb06578b0a2f272869
  vector 00 [ascii-alnum-16      ] A=PASS B=PASS
  vector 01 [empty-password      ] A=PASS B=PASS
  vector 02 [long-password-1200  ] A=PASS B=PASS
  vector 03 [utf8-combining      ] A=PASS B=PASS
  vector 04 [high-bytes-utf8     ] A=PASS B=PASS
  vector 05 [shell-meta          ] A=PASS B=PASS
  overall: PASS
  ```

  This is the strongest local proof available: byte-for-byte
  agreement between btt's rust `argon2i + xsalsa20poly1305` path and
  pynacl + argon2-cffi, across all six vectors, in both directions.

## Test plan

- [ ] CI: btcli-compat workflow runs and all six vectors pass in
      both directions under pinned `--require-hashes` deps
- [ ] CI: cargo build + cargo test on ubuntu-latest (this PR
      restores the build)
- [ ] Download the `btcli-compat-vectors-<run_id>` artifact from the
      successful run and spot-check one vector by decrypting the
      `vector-00-btt.bin` blob manually with pynacl
- [ ] Manual: `btt wallet create --name foo --password-file /dev/shm/pw`
      with a mode-0600 file, then without, then with a mode-0644
      file (should refuse with "insecure mode")
- [ ] Manual: confirm `btt wallet sign --help` shows the full
      security warning paragraph in the --password-file help text
- [ ] Manual: confirm `btt skill` output documents `--password-file`
      under every affected command

## Notes / deviations

- **`main` was already broken.** The `wallet-keys` merge introduced
  `print_success(..., pretty, quiet)` three-arg calls into `main.rs`
  after the earlier `fix(output)` commit had reverted `print_success`
  to two args. `cargo build` failed with 17 errors on `origin/main`
  as of this PR's base. This PR restores the build as a necessary
  side effect — the btcli-compat workflow can't run on a broken
  build. I opted for a single commit covering both the flag
  plumbing and the three-arg fix because the new match arms would
  overwrite those lines anyway; splitting them would have been
  cosmetic.
- **Bootstrap re-pack approach for direction B.** Rather than
  hand-constructing the inner keyfile JSON (which would couple the
  verifier to btt's internal JSON shape), direction B asks btt to
  produce a bootstrap blob, decrypts it with pynacl to extract the
  inner JSON, and re-encrypts *that same JSON* under the vector
  password. btt is then asked to sign. The envelope is the only
  thing we're proving interoperability on; the inner format is
  btt's own.
- **rust-toolchain.toml does not exist in this repo**, so the
  workflow uses `dtolnay/rust-toolchain@stable` rather than pinning
  a specific nightly/stable. If a future commit adds a
  rust-toolchain file, the workflow action will pick it up
  automatically.
- **No new direct rust deps.** `zeroize` was already present; I use
  its `Zeroizing` wrapper only. No new entries in `Cargo.toml` /
  `Cargo.lock`.
- **No `bittensor` / `btwallet` in the pin file** and no plan to
  ever add one. The pin file comment makes this explicit.

## Open questions

None blocking. A couple of things worth flagging for review:

- The workflow uses `dtolnay/rust-toolchain@stable`. If the project
  wants a specific pinned rust version for CI reproducibility, that
  belongs in a `rust-toolchain.toml` (outside this PR's scope).
- The pinned python deps are going to need re-pinning when cffi or
  argon2-cffi cut a new release. A follow-up task — probably
  triggered by the weekly cron starting to fail — will handle that.

### DIFF
--- a/.github/workflows/btcli-compat.yml
+++ b/.github/workflows/btcli-compat.yml
@@ -1,0 +1,137 @@
+name: btcli-compat
+
+# Byte-for-byte verification that btt's coldkey envelope is wire-compatible
+# with btcli/btwallet, using pinned pynacl + argon2-cffi primitives only.
+# See scripts/btcli-compat/README.md for the design.
+
+on:
+  pull_request:
+    paths:
+      - "src/commands/wallet_keys.rs"
+      - "src/commands/password_file.rs"
+      - "Cargo.toml"
+      - "Cargo.lock"
+      - "scripts/btcli-compat/**"
+      - ".github/workflows/btcli-compat.yml"
+  push:
+    branches:
+      - main
+  schedule:
+    # Weekly drift detector. If pynacl or argon2-cffi quietly change
+    # semantics in a way our pinned hashes somehow allow, this is how
+    # we find out.
+    - cron: "0 3 * * 0"
+  workflow_dispatch:
+    inputs:
+      btcli_version:
+        description: "Ignored — this workflow never installs btcli. Reserved for future cross-validation."
+        required: false
+        type: string
+      vector_count:
+        description: "Number of compat vectors (max 6)"
+        required: false
+        default: "6"
+        type: string
+      verbose:
+        description: "Echo every btt invocation to stderr"
+        required: false
+        default: false
+        type: boolean
+
+permissions:
+  contents: read
+
+jobs:
+  verify:
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    env:
+      CARGO_TERM_COLOR: always
+      # Be explicit about which python the workflow gets. A future
+      # actions/setup-python bump to 3.12 or 3.13 needs a matching
+      # pin file audit.
+      BTCLI_COMPAT_PY: "3.11"
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Setup Rust toolchain
+        uses: dtolnay/rust-toolchain@stable
+        with:
+          components: clippy
+
+      - name: Cargo cache
+        uses: actions/cache@v4
+        with:
+          path: |
+            ~/.cargo/bin/
+            ~/.cargo/registry/index/
+            ~/.cargo/registry/cache/
+            ~/.cargo/git/db/
+            target/
+          key: ${{ runner.os }}-cargo-btcli-compat-${{ hashFiles('**/Cargo.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-cargo-btcli-compat-
+
+      - name: Build btt (release)
+        run: cargo build --release
+
+      - name: Setup Python ${{ env.BTCLI_COMPAT_PY }}
+        uses: actions/setup-python@v5
+        with:
+          python-version: ${{ env.BTCLI_COMPAT_PY }}
+
+      - name: Install pinned verifier deps
+        # --require-hashes forces every dep (transitive included) to match
+        # a hash in the pin file. If a transitive changes, the install
+        # fails and CI fails, which is the intended behaviour.
+        run: |
+          python -m pip install --upgrade pip
+          python -m pip install --require-hashes \
+              -r scripts/btcli-compat/btcli-pins.txt
+          python -c "import nacl; import argon2; print('nacl', nacl.__version__); print('argon2 ok')"
+
+      - name: Run compat check
+        env:
+          VECTOR_COUNT: ${{ github.event.inputs.vector_count || '6' }}
+          VERBOSE: ${{ github.event.inputs.verbose || 'false' }}
+        run: |
+          set -euo pipefail
+          extra_args=""
+          if [ "${VERBOSE}" = "true" ]; then
+              extra_args="--verbose"
+          fi
+          python scripts/btcli-compat/check.py \
+              --btt-binary ./target/release/btt \
+              --out-dir ./btcli-compat-out \
+              --vector-count "${VECTOR_COUNT}" \
+              ${extra_args}
+
+      - name: Upload compat vectors and report
+        if: always()
+        uses: actions/upload-artifact@v4
+        with:
+          name: btcli-compat-vectors-${{ github.run_id }}
+          path: btcli-compat-out/
+          retention-days: 90
+          if-no-files-found: warn
+
+      - name: Summarise
+        if: always()
+        run: |
+          if [ -f btcli-compat-out/report.json ]; then
+              python - <<'PY'
+          import json, pathlib, sys
+          rep = json.loads(pathlib.Path("btcli-compat-out/report.json").read_text())
+          print(f"overall: {rep['overall']}")
+          print(f"seed:    {rep['seed']}")
+          for r in rep["results"]:
+              print(
+                  f"  v{r['vector_index']:02d} [{r['label']:<22}] "
+                  f"A={r['btt_to_pynacl']:<4} B={r['pynacl_to_btt']:<4} "
+                  f"{r.get('error') or ''}"
+              )
+          PY
+          else
+              echo "no report.json produced; check.py crashed"
+          fi
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -47,6 +47,8 @@ This is not a style preference. Agents and scripts must be able to parse `btt` o
 
 Documentation lies. Comments lie. Memoized state lies. When we need to know what the chain says, we query the chain. When we need to know what a contract or runtime does, we read the bytecode or the source. Trust boundaries are drawn around the Substrate RPC connection, not around anything upstream of it.
 
+btcli keyfile format compatibility is verified in CI by `scripts/btcli-compat/check.py` against pinned `pynacl` and `argon2-cffi` primitives, never against `bittensor.*`. The verifier lives only in CI runners and never in dev environments or release artifacts.
+
 ### 5. Barbaric review
 
 Every PR receives one round of hostile review before merge. Reviewers are explicitly instructed to find what is wrong with the change, not to confirm what is right. Findings are classified by severity (CRITICAL, HIGH, MEDIUM, LOW, NIT). Any HIGH or CRITICAL finding blocks merge.
--- a/README.md
+++ b/README.md
@@ -25,3 +25,30 @@ btt skill
 ```
 
 All commands output JSON to stdout. Use `--pretty` for human-readable formatting.
+
+## Non-interactive automation
+
+Commands that prompt for the coldkey password (`wallet create`, `wallet
+new-coldkey`, `wallet regen-coldkey`, `wallet sign`) accept a `--password-file
+<path>` flag for CI and scripted use. The file's first line (minus the
+trailing newline) is taken as the password.
+
+```bash
+# Prepare the password on a tmpfs at mode 0600.
+umask 077
+printf '%s\n' "$BTT_COLDKEY_PW" > /dev/shm/btt-pw
+btt wallet create --name compat-test --password-file /dev/shm/btt-pw
+shred -u /dev/shm/btt-pw
+```
+
+On unix, btt refuses to read the file if its mode is other-readable. Do not
+use `--password-file` with mainnet wallets unless your filesystem, process
+listing, and shell history are all under your control.
+
+## btcli format compatibility
+
+btt's coldkey envelope (`$NACL` + argon2i13 SENSITIVE + xsalsa20poly1305) is
+verified byte-for-byte against the reference primitives (pynacl,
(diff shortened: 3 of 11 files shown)


moorkh:
[cryptid] barbaric review of PR #12

## Prior design constraints — verdict per item

| # | Constraint | Status | Notes |
|---|---|---|---|
| 1 | `bittensor.*` never imported in `check.py` | PASS | No `import bittensor`, `btwallet`, `substrate-interface`, or `scalecodec` anywhere in the script. Only `nacl.*` and `argon2.low_level`. |
| 2 | No hidden CLI flags | PASS | No `hide = true` on any `--password-file` arg. `--help`, `btt skill` output, and README all advertise the flag. |
| 3 | `btcli-pins.txt` uses `--require-hashes` with real sha256 hashes | PASS | Every one of the 15 pinned hashes across pynacl 1.6.2, argon2-cffi 25.1.0, argon2-cffi-bindings 25.1.0, cffi 2.0.0, pycparser 3.0 was fetched from `https://pypi.org/pypi/<pkg>/<ver>/json` and matched a real artifact. No placeholders. |
| 4 | Argon2i13 SENSITIVE params match libsodium | PASS | `ARGON2_T_COST=8`, `ARGON2_M_COST_KIB=524288`, `ARGON2_PARALLELISM=1`, `hash_len=32`, `type=ArgonType.I`. Matches `wallet_keys.rs` exactly (t_cost=8, m_cost=524288, p=1, Argon2i, V0x13). Not argon2id. |
| 5 | `NACL_SALT` bytes match Rust source | PASS | Python: `bytes.fromhex(\"137183dff15a09bc9c90b5518739e9b1\")`. Rust: `b\"\\x13q\\x83\\xdf\\xf1Z\\x09\\xbc\\x9c\\x90\\xb5Q\\x879\\xe9\\xb1\"` → same 16 bytes. |
| 6 | `--password-file` security checks | PASS | `mode & 0o077 != 0` rejection, `is_file()` rejection of FIFO/device/dir, first-line + `\\r\\n`-strip, `expand_tilde`, `Zeroizing<String>` return type, 64 KiB cap, clear error messages. 13 unit tests cover the lot and all pass. |
| 7 | No new btt direct deps | PASS | `Cargo.toml` unchanged. Uses the existing `zeroize` + `argon2` already on main. |
| 8 | Flag absent → interactive path unchanged | PASS | `resolve_coldkey_password(None)` falls through to the pre-existing `read_password()` rpassword prompt. |
| 9 | All 4 subcommands accept `--password-file` | PASS | `wallet create`, `wallet new-coldkey`, `wallet regen-coldkey`, `wallet sign`. Verified at the clap level and in `main.rs`. |
| 10 | Build-break fix in `main.rs` is correct | PASS | `output::print_success` was refactored to 2 args in PR #3 but stale 3-arg sites remained on `origin/main`. PR drops the stale `quiet` / `false` third arg at every call site. Not papering — it's the correct follow-through of the earlier refactor. |

## Local run — actually executed

- `cargo build --release` on 0906e70: success.
- `cargo test --release`: **67 passed, 0 failed, 1 ignored** (the ignored test is the pre-existing `dump_blob_for_python` helper). All 13 `password_file` tests pass including `rejects_fifo`, `rejects_group_readable`, `rejects_world_readable`, `strips_crlf`, `ignores_content_past_first_newline`, `rejects_directory`.
- `python3 scripts/btcli-compat/check.py --btt-binary ./target/release/btt --vector-count 6`: **overall PASS**, exit 0. All 6 vectors pass direction A (btt→pynacl) and direction B (pynacl→btt). Labels: `ascii-alnum-16`, `empty-password`, `long-password-1200`, `utf8-combining`, `high-bytes-utf8`, `shell-meta`.

The envelope-compat claim is not speculation. I watched pynacl decrypt what btt wrote, and I watched btt sign with a keyfile that pynacl re-enveloped. Byte for byte.

## New findings

### MEDIUM — TOCTOU between permission check and file open in `password_file.rs`

`read_password_file_inner` calls `fs::metadata(path)` to stat the file (which follows symlinks), then later calls `fs::File::open(path)`. Between those two syscalls, an attacker with write access to any directory on the path can swap the target — e.g. point a symlink at a mode-0600 decoy during the stat, then at a mode-0644 or attacker-controlled file during the open. The current code will then happily read whatever is at the swapped target.

In practice the exposure is narrow: the threat model already assumes the operator controls the password-file's directory, and the current callers all use a fresh tmpfs path the user just wrote to. Still, the check is cheap and the fix obvious:

```rust
let file = fs::File::open(path)?;        // open first
let md   = file.metadata()?;              // stat the fd, not the path
// then is_file / mode checks on `md`
```

This closes the race because `File::metadata` operates on the open file descriptor — whatever it returns IS what you just opened, not a racy re-lookup. It also lets you drop the separate `fs::metadata` call entirely.

Non-blocking because the attacker model here is narrow, but should be fixed as a follow-up.

### LOW — `fs::metadata` follows symlinks, so `is_file()` can be tricked post-open

Related but distinct: even without a race, `metadata()` follows symlinks, so a symlink pointing at `/dev/zero` or a FIFO would get the metadata of the target. That IS what the current code wants (it blocks the FIFO), so this isn't a bug — but combined with the TOCTOU above, the symlink-follow makes the race trivially weaponizable if an attacker can control any component of the path. The fd-based fix above resolves both at once.

### LOW — Workflow `paths:` filter misses `src/main.rs` and `src/cli.rs`

`.github/workflows/btcli-compat.yml` triggers on changes to `src/commands/wallet_keys.rs`, `src/commands/password_file.rs`, `Cargo.toml`, `Cargo.lock`, and `scripts/btcli-compat/**`. It does NOT trigger on changes to `src/main.rs` or `src/cli.rs`. A future PR that rewires the `--password-file` plumbing in `main.rs` (or adds a new coldkey-touching subcommand in `cli.rs`) without touching `wallet_keys.rs` would silently skip this workflow. Add both files to the filter.

### LOW — `std::env::set_var` in a `#[test]` without serial ordering

`password_file::tests::expand_tilde_prefix` calls `std::env::set_var(\"HOME\", ...)` from a unit test. Cargo runs tests in parallel by default, and `set_var` is process-global and marked `unsafe` in Rust 1.82+. If any future test reads `HOME`, it will race. Not triggering today, but a footgun. Either gate with `#[serial]` (needs the `serial_test` dep — declined under the dep-discipline rule) or, better, refactor `expand_tilde` to take the home dir as an argument and test it pure.

### INFO — `dtolnay/rust-toolchain@stable` is not SHA-pinned

The workflow uses `actions/checkout@v4`, `actions/setup-python@v5`, `actions/cache@v4`, `actions/upload-artifact@v4` — all tag-pinned but published by GitHub. `dtolnay/rust-toolchain@stable` is the odd one out: third-party, pinned to a moving `stable` tag. It's the de-facto community standard and replacing it would add friction, so this is informational, not a finding. If you ever harden the whole workflow set to SHA pins, add it to the list.

### INFO — `check.py` shred fallback can fail silently on a read-only filesystem

`shred_and_remove` falls through a chain and swallows `OSError`. On a tmpfs with a weird mount option, the overwrite can silently no-op. Since the caller still calls `path.unlink(missing_ok=True)` at the end, the file does get removed — but the \"shredded\" claim becomes symbolic. Not exploitable, already flagged in the docstring (\"A tmpfs makes this mostly symbolic\"), just noting it.

## Verdict

**APPROVE.**

Every prior design constraint is met. The cryptographic core is correct — I verified the salt, the Argon2 parameters, and the envelope format against libsodium-backed primitives, and I watched the bidirectional compat pass on 6 vectors including empty-password, 1200-byte, combining-mark UTF-8, and shell-metacharacter passwords. The password-handling code is `Zeroizing` end to end, there are zero `.unwrap()` calls in production paths, the 13 unit tests cover the interesting failure modes (world-readable, group-readable, FIFO, directory, missing, CRLF, no trailing newline, content past first newline, tilde expansion). The pip pin file is real — every hash resolves on pypi. The workflow uploads artifacts on failure and runs weekly as a drift detector.

Findings are:
- 1 MEDIUM (TOCTOU on password-file open — non-blocking, narrow threat model, fix is a three-line refactor)
- 3 LOW
- 2 INFO

Track MEDIUM + LOWs as follow-ups. Nothing here blocks merge.

The scar tissue from July 2024 is visible in this PR and it is the right scar tissue. The verifier never imports `bittensor`. The envelope is the feature. The check IS the product. Merge it.

Issue #15, opened 2026-04-13
ci(dep-audit): parallelize via matrix and fix cache key


moorkh:
[cryptid]

Follow-up to PR #10 (now merged at `5a8fb5f3`). Two related improvements to the dep-audit workflow:

## 1. Parallelize the three tools via matrix strategy

Currently `cargo-audit`, `cargo-deny`, and `cargo-outdated` run sequentially in one job. Wall time on the warm-cache path is dominated by their cumulative run time. Switch to a matrix strategy:

```yaml
jobs:
  audit:
    strategy:
      matrix:
        tool: [cargo-audit, cargo-deny, cargo-outdated]
    steps:
      - name: Run ${{ matrix.tool }}
        run: scripts/dep-audit/run-one.sh ${{ matrix.tool }}
      - name: Upload artifact
        uses: actions/upload-artifact@v4
        with:
          name: dep-audit-${{ matrix.tool }}-${{ github.run_id }}
          path: report.md
  combine:
    needs: [audit]
    if: always()
    steps:
      - name: Download all reports
      - name: Combine into one Markdown
      - name: Post or update PR comment
```

Wall time becomes `max(audit, deny, outdated) + combine` instead of `audit + deny + outdated + post`. Saves ~50% on warm-cache runs.

The combine job posts a single combined PR comment using the same `<!-- cryptid-dep-audit -->` marker, so we don't end up with three separate comments cluttering the PR.

The `scripts/dep-audit/run-one.sh` wrapper is small — extracts the per-tool logic from the current monolithic `audit.sh` and runs only one section.

## 2. Fix the cache key to invalidate on tool updates

Currently:

```yaml
key: cargo-audit-tools-v1
```

This key never invalidates. When cargo-audit ships a new version with detection for a new advisory class, the cache hit serves the old binary that doesn't know about it.

Fix:

```yaml
key: cargo-audit-tools-${{ runner.os }}-${{ hashFiles('scripts/dep-audit/tool-versions.txt') }}
```

Where `scripts/dep-audit/tool-versions.txt` pins the versions we want:

```
cargo-audit==0.21.2
cargo-deny==0.16.4
cargo-outdated==0.16.0
```

The install step uses `cargo install --locked cargo-audit@<version>` etc, reading from the same file. When we want to update a tool, we bump the version in `tool-versions.txt` and the next run installs fresh (and caches the new binary under a new key).

## Side benefits

[...]

PR #16, opened 2026-04-13
ci(dep-audit): parallelize via matrix and fix cache key

moorkh:
[cryptid]

Closes #15.

## Summary

Splits the dep-audit workflow into a three-cell matrix (`cargo-audit`, `cargo-deny`, `cargo-outdated`) that runs in parallel, plus a `combine` job that concatenates the per-tool reports into a single PR comment under the existing `<!-- cryptid-dep-audit -->` marker. Also pins tool versions in `scripts/dep-audit/tool-versions.txt` and rewrites the cache key to hash that file so bumps invalidate cleanly.

Wall time on the warm-cache path becomes `max(audit, deny, outdated) + combine` instead of the roughly ten-minute sequential sum.

## Acceptance criteria

- [x] Three tools run in parallel matrix jobs
- [x] Combine job posts a single PR comment with all three reports
- [x] Cache key includes a hash of the tool versions file
- [x] Tool versions are pinned in `tool-versions.txt` and the install commands read from it
- [x] Wall time on warm-cache run is < 5 min — cold-cache run 24351199503 measured at ~4m56s wall (max matrix leg 286s + combine 10s); warm cache will skip the `cargo install` step entirely once the cache scope populates post-merge
- [x] CI evidence: successful run linked below

## Files changed

- `scripts/dep-audit/tool-versions.txt` (new) — pins `cargo-audit = 0.22.1`, `cargo-deny = 0.19.1`, `cargo-outdated = 0.18.0` as the single source of truth for both the cache-key hash and the install command version.
- `scripts/dep-audit/run-one.sh` (new) — takes one of the three tool names and emits that tool's Markdown section on stdout, exiting with the tool's blocking status. Shared color-suppression env, shared repo-root detection, shared advisory ignore list.
- `scripts/dep-audit/audit.sh` — reduced to a thin wrapper that chains `run-one.sh cargo-audit && run-one.sh cargo-deny && run-one.sh cargo-outdated`, writes the `## Dep audit` header and `STATUS: PASS|FAIL` footer, and returns the union exit code. Preserved for local use.
- `.github/workflows/dep-audit.yml` — restructured into:
[...]

PR #17, opened 2026-04-13
cryptid/dep-audit: dep-checksum tripwire backed by a committed DB

### LINKED ISSUE #6: cryptid/dep-audit: dependency checksum tripwire backed by ken

moorkh:
Closes #6.

## What this adds

A supply-chain tripwire that detects the class of attack cargo-audit and cargo-deny cannot: same dep, same version, different bytes. This is the failure mode that destroyed btcli in 2024 — a maintainer-trust republish rather than a discoverable CVE.

`scripts/dep-audit/checksum.py` parses every `[[package]]` block in `Cargo.lock`, computes a canonical sha256 per dep, and compares against `.cryptid/dep-audit/checksums.db`. The DB is committed into the repo and its git history is the audit log. Same script runs in CI and locally.

## Files changed

- `scripts/dep-audit/checksum.py` (new, 356 lines, executable) — the tripwire.
- `.cryptid/dep-audit/checksums.db` (new, 972 entries) — seeded baseline from the current `Cargo.lock`.
- `.github/workflows/dep-audit.yml` — new step after the existing audit run; appends the checksum report to `audit-report.md` and flips STATUS to FAIL on a tripwire hit.
- `scripts/dep-audit/README.md` — new section documenting the tripwire, the DB format, the source-kind checksum semantics, the `--update` flag, and a runbook for a CRITICAL finding.

## Design choices (resolving the open questions from the issue)

**Git-source checksum mechanism: the commit SHA itself.** The refinement comment suggested cargo-vendored output sha256 as first preference. On reflection, using the commit SHA encoded in the `Cargo.lock` source URL fragment is strictly better: git commits are already content-addressed over the tree, the SHA is already in `Cargo.lock`, and it requires zero subprocess invocation, zero network, and zero tool dependency. The security property is identical — a moved git ref is exactly a commit SHA change. btt has no git-sourced deps today but this code path is in place for when it does.

**Registry-source checksum: the sha256 cargo already wrote into `Cargo.lock`.** Cargo stores the .crate tarball hash in the `checksum` field of every registry package and verifies it at fetch time. Reusing that value means the tripwire sees exactly what cargo sees — same bytes cargo would refuse to build with if the registry served different content. Zero network, zero new trust assumptions.

**Path-source checksum: deterministic manifest sha256.** For path deps, the script walks the tree, emits `(kind, relpath, sha256)` tuples, sorts, and hashes the concatenation. btt has no path deps today but the tripwire catches a path-override injection if one is ever added.

**Script language: pure Python 3 stdlib.** Cleaner than bash for Cargo.lock parsing, and the cryptid host constraint (no new local deps, see `feedback_no_local_verification_deps.md`) rules out anything requiring pip. Python 3 is already on `ubuntu-latest`. No `tomllib` dependency — the parser handles just the `[[package]]` blocks it needs.

**DB serialization: plain text, tab-delimited, lexicographically sorted.** Format is `name@version\tsource-kind\tsha256`. The simplest thing that gives a readable `git diff`, a trivial operator-side grep/sort surface, and zero parser dependencies. A TOML serialization would add a parser, a YAML serialization would introduce quoting complexity; neither renders as cleanly in a PR diff as one-line-per-entry text.

**Workflow integration: extend `dep-audit.yml`, not a new workflow.** Same concern (dep integrity), same triggers, keeps the PR comment surface small. Couples to PR #15 (parallelization) — may need a rebase if that lands first.

## Test plan

- [x] `python3 -m py_compile scripts/dep-audit/checksum.py` — clean
- [x] `python3 -c "import yaml; yaml.safe_load(open('.github/workflows/dep-audit.yml'))"` — clean
- [x] Seeding run: `python3 scripts/dep-audit/checksum.py --update` produces 972 DB entries (exactly `wc -l` of `[[package]]` blocks in Cargo.lock minus the 1 workspace crate)
- [x] Verification run: `python3 scripts/dep-audit/checksum.py` on the seeded DB exits 0 with `STATUS: PASS`, `matches: 972, new: 0, mismatches: 0`
- [x] Mismatch detection: flipping one hex char in one DB entry causes the script to exit 1, print `CRITICAL — checksum mismatch on existing entry`, and name the dep with both hashes side by side
- [x] New-entry-without-update: removing one line from the DB causes the script to exit 1 and list the missing dep under `New deps (not yet in DB)`
- [ ] CI run on this PR (will be visible once workflows execute)

## CI evidence

Will be populated by the dep-audit workflow run on this PR. The PR comment posted by the workflow should include a new `### dep-checksum tripwire` section after the cargo-audit/deny/outdated sections, reading `Tracked: 972 deps; DB entries: 972; matches: 972; new: 0; stale: 0; mismatches: 0.`

## Operator notes

- Never run `--update` to silence a CRITICAL finding. That disarms the tripwire. `scripts/dep-audit/README.md` has the runbook.
- CI never passes `--update`. Any Cargo.lock change that does not also update the DB fails the PR — which is the point.
- The DB's git history is the audit log. `git log .cryptid/dep-audit/` shows every dep-byte change with an author and a commit message. Reverts are `git revert`.


moorkh:
[cryptid] barbaric review of PR #17 (round 1, re-dispatch after API outage)

## Verdict

**APPROVE** with two low-severity findings filed as follow-ups. Every required check in the scope passes. The script's three test flows (match / mismatch / new) reproduce exactly as advertised and the committed DB is byte-stable.

Severity: 0 CRITICAL · 0 HIGH · 0 MEDIUM · 2 LOW · 2 INFO

## What I verified

### 1. Cargo.lock parser correctness — PASS
Regex parser `^(\w+)\s*=\s*"(.*)"$` correctly ignores:
- `dependencies = [` list openers and ` "rand 0.8.5",` continuation lines (leading whitespace, no `\w+` prefix)
- Workspace crate block with no `source` field (btt@0.1.0, Cargo.lock L1389–1405) — classified as `workspace`, skipped
- Trailing `[[metadata]]` sections (v1 legacy format) — `line.startswith("[")` with `!= "[[package]]"` terminates the in_package state
- v3 vs v4 lockfiles — the top-level `version = N` key sits outside any `[[package]]` and is ignored
- Adjacent blocks with no blank line — the `[[package]]` header appends the prior `current` before starting new
- Duplicate `name@version` with divergent source/sha — explicitly caught, exits 2

Real lockfile: 973 `[[package]]` blocks → 972 tracked (one workspace crate skipped) → 972 DB entries. Clean.

### 2. Registry-source checksum reuse — PASS
`checksum_registry` reads `pkg["checksum"]` directly from the parsed block. No hashing, no fetch, no re-computation. This is the single strongest design choice in the PR: those bytes are exactly what cargo refuses to build against if the registry serves different content. Zero new trust surface.

### 3. Git-source fragment extraction — PASS
`rsplit("#", 1)[1]` cleanly separates the commit fragment from the URL, so multiple `?` query parameters in the source (e.g. `git+https://host/repo?branch=main&rev=abc#<sha>`) are irrelevant — the split is on `#`, not `?`. Validated with synthetic `git+https://github.com/foo/bar?rev=abc&branch=main#1234...5678` → extracts `1234...5678`. Missing fragment raises `error: git dep ... has no commit fragment`. Non-sha1 fragment raises `error: ... has non-sha1 commit fragment`. Both hard-fail.

### 4. Path-source manifest — PASS (with one latent bug, LOW)
Manifest includes every regular file and symlink under the path dep root, recursively; walk order is stabilised by `dirnames.sort()` plus `sorted(filenames)`. Each entry is `f|l <relpath> <sha256>`. Symlinks hash the link target string, not its resolved content. Everything inside the dep root is included — there is no exclusion of `target/`, `Cargo.lock`, or hidden files. Since btt has zero path deps today this is cold code, but worth pinning down before a future path-dep lands (see Finding 1 below).

### 5. DB byte-stability across runs — PASS
Reseeded the DB twice from scratch via `--update` after deleting the existing file. Both runs produced identical sha256 `c9d0c20...c2a8750`, identical to the committed DB. Python's `str.sort()` is codepoint-based (locale-independent); capitals sort before lowercase (`Inflector@0.11.4` precedes `addr2line@0.19.0`), exactly as on disk.

### 6. Three test states reproduced — PASS
- **match**: unmodified DB → `matches: 972, new: 0, mismatches: 0`, STATUS: PASS, exit 0
- **mismatch**: flipped the last hex char of `Inflector@0.11.4` → `mismatches: 1`, CRITICAL section prints both hashes side by side, STATUS: FAIL, exit 1
- **new**: removed the `Inflector@0.11.4` line from the DB → `new: 1`, prints the missing dep under "New deps (not yet in DB)", STATUS: FAIL, exit 1

### 7. Pure stdlib — PASS
`python3 -m py_compile` clean. Imports: `argparse`, `hashlib`, `os`, `re`, `sys`, `pathlib.Path`. Zero third-party. No `tomllib` dependency. Runnable on any Python 3.

### 8. Workflow integration — PASS
New step "Run dep-checksum tripwire" runs after the cargo-audit/deny/outdated dispatcher, before the PR-comment step. `set +e` around the invocation captures the exit into `checksum_exit` without aborting the job. Tripwire FAIL rewrites `STATUS: PASS` → `STATUS: FAIL` via `sed -i` before the final "Fail job on critical findings" step re-greps STATUS. `--update` appears only in a code comment, never as an actual flag. The `checksum.py` invocation is free of `--update`. `checksum-report.md`, `checksum-stdout.log`, `checksum-stderr.log` are all uploaded as artifacts for 90-day retention. Couples to PR #16 (parallelization) as the PR body warns; if #16 lands first this will need a rebase to slot the new step into the right parallel leg.

### 9. CLI surface — PASS
`--help` documents `--update`, `--lock`, `--db`, `--report`. Every flag on the parser is discoverable. No hidden surface. Complies with `feedback_no_hidden_flags.md`.

### 10. Coordinated-tampering review surface — PASS
A single hex-char flip produces a single-line `diff -u` (verified). Line-oriented, sorted, tab-delimited, one-entry-per-line format renders readably in a GitHub PR diff. An attacker writing both `Cargo.lock` and `checksums.db` in the same PR leaves a visible DB-diff for the reviewer to inspect against the Cargo.lock delta. The DB header explicitly warns `Do not edit by hand` and points at the `--update` workflow, so a hand-edited DB looks wrong at a glance (wrong sort, trailing whitespace, missing header).

## Findings

### LOW-1 · `path+file://` source URL handling is broken (latent)
`classify_source` strips the `path+` prefix leaving `file:///abs/path`; `checksum_path` then does `Path("file:///abs/path")`, which is not absolute on POSIX, so it gets joined with `REPO_ROOT` → a nonexistent path → hard fail. Synthetic repro:

```
source = "path+file:///tmp/cksum-test/pathdep"
→ error: path dep ... points at /tmp/btt-pr17-review/file:/tmp/cksum-test/pathdep which does not exist
```

Impact: zero today because btt has no path deps — the real lockfile contains zero `path+` sources. But the moment a path dep lands, this breaks the tripwire. The fix is two lines: strip a leading `file://` (or `file:`) prefix before constructing the `Path`. File as a follow-up issue and link it from the `checksum_path` docstring so the next editor notices before landing a path dep.

### LOW-2 · Path-dep manifest has no exclusion rules documented or enforced
Every file under the path-dep root is hashed, including `target/`, `.git/`, editor swap files, OS metadata, and anything a build may have left behind. For a future path dep inside a workspace that has its own `target/`, the manifest will include gigabytes of build artefacts and change on every local compile, making the tripwire noisy-to-useless for that dep. Either (a) hard-exclude `target/`, `.git/`, `*.swp`, `.DS_Store`, or (b) pin the manifest to only files tracked by `git ls-files` for that subtree. Not blocking because no path deps exist, but worth pinning down in the same follow-up as LOW-1.

### INFO-1 · Exit code contract vs docstring
The docstring promises exit 2 for script errors (unreadable Cargo.lock, unknown source kind). In practice, `raise SystemExit(f"error: ...")` with a string message exits 1, not 2. Two call sites affected: `checksum_registry` on a missing checksum field, `checksum_git` on missing/invalid fragment, `classify_source` on unknown source kind. Only `compute_entry`'s duplicate-package check uses the explicit `return 2` path. Low-impact because the CI step doesn't distinguish exit 1 from exit 2 — both fail the job — but the docstring is inaccurate. Either change the `SystemExit("error: ...")` calls to `raise SystemExit(2)` with a printed message, or update the docstring to say "exit 1 on any error". I would not block merge on this.

### INFO-2 · Cosmetic report contradiction during `--update`
On a `--update` run that has new entries, the report still prints `**New deps (not yet in DB):**` and `These entries are not in the committed DB. Run ... --update ...` even though the script is in the middle of writing them. The final `STATUS: PASS` is correct, but the interstitial messaging is confusing. Suppress the "Run --update" paragraph when `args.update` is true. Purely cosmetic.

## What I liked

The reuse of cargo's `checksum` field is the point of the whole design. It means btt's tripwire and cargo's own registry verification are looking at the same bytes, so a supply-chain swap that evades the tripwire also evades cargo — and cargo refuses to build in that case. There is no gap. That is a beautiful property and worth keeping central to any future refactor.

The decision to use the git commit SHA directly, rather than fetching the repo and vendoring it, is similarly elegant. Git commits are already content-addressed over the tree; the SHA *is* the canonical content hash. Zero network, zero tooling, same security property. I note your PR body pushes back on the earlier "cargo-vendored output" suggestion — your instinct was right.

Line-oriented sorted text for the DB is exactly the right serialization for a git-history-as-audit-log. Any richer format would obscure the diff.

## Housekeeping

All artefacts cleaned up. Worktree was reused from the pre-crash attempt and is ready for removal.

PR #18, opened 2026-04-13
fix(wallet): refuse to overwrite existing keys without --force

### LINKED ISSUE #7: btt wallet new-coldkey/new-hotkey silently overwrites existing keys
[cryptid]

## Summary

`btt wallet new-coldkey` and `btt wallet new-hotkey` will silently destroy an existing key file at the target path. The original ss58 is replaced without prompt or confirmation. Discovered during the round-2 barbaric review of PR #3 (NEW-M1).

## Severity

MEDIUM. Not a code-correctness bug — the encryption is correct, the writes are atomic, and the new file is at 0600. The footgun is human-facing: the user invokes `btt wallet new-coldkey --name foo` thinking they are creating a new wallet, and instead discovers that an existing wallet at `~/.bittensor/wallets/foo` has been overwritten and is now unrecoverable unless they have the original mnemonic.

This is a data-loss path on the most dangerous commands in the CLI. btcli prompts here; btt currently does not.

## Mechanism

`write_secure_file` (in `src/commands/wallet_keys.rs`) is invoked unconditionally, and the underlying `OpenOptions::create_new(true)` will *fail* if the file exists — but the calling code in `new_coldkey` / `new_hotkey` removes the existing file before calling `write_secure_file`. The removal step is what makes the operation destructive without prompt.

## Reproduction

```
$ HOME=/tmp/btt-wipe-test cargo run --release -- wallet new-coldkey --name foo
[prompts for password, prints original ss58 = 5Abc...]
$ HOME=/tmp/btt-wipe-test cargo run --release -- wallet new-coldkey --name foo
[prompts for password, prints DIFFERENT ss58 = 5Xyz...]
$ # the original 5Abc... mnemonic is gone forever
```

## Fix

Two acceptable shapes; prefer (1) for safety, fall back to (2) only if (1) is judged too friction-heavy.

1. **Refuse to overwrite by default. Require an explicit `--force` flag.** `new-coldkey` / `new-hotkey` check if the target path exists; if yes, error with a message naming the existing file and instructing the user to either delete it manually or pass `--force`. This is the fail-closed shape and matches the security-first ethos of the project.
2. **Prompt interactively if the target exists**, with the existing ss58 displayed and a y/N confirmation. This matches btcli's behavior. Less safe in non-interactive contexts (the prompt would have to error rather than default-yes).

I lean toward (1). The transparency principle from no-hidden-flags applies: the user types `--force`, the user knows what they did. No interactive surprise either way.

The same fix should apply to `new-coldkey`, `new-hotkey`, `regen-coldkey`, `regen-hotkey`. `create` already errors on existing wallets per the create_new semantics — confirm and document.

## Acceptance criteria

- [ ] `new-coldkey`, `new-hotkey`, `regen-coldkey`, `regen-hotkey` all refuse to overwrite by default
- [ ] `--force` flag added to each, documented in --help, in `btt skill`, in README automation section (per no-hidden-flags principle)
- [ ] Test that reproduces the original footgun and confirms the fixed version errors
- [ ] Test that confirms `--force` does the overwrite intentionally
- [ ] AGENTS.md does not need updating (this is a CLI hygiene point, not a principle)

## Related

- Discovered in PR #3 round-2 review: https://github.com/ErgodicLabs/btt/pull/3#issuecomment-4236598774
- Severity per AGENTS.md principle 5: MEDIUM, does not block merge of PR #3

moorkh:
Closes #7.

## Summary

Before this PR, `btt wallet new-coldkey`, `new-hotkey`, `regen-coldkey`, and `regen-hotkey` silently destroyed any existing key file at the target path — a second invocation of `new-coldkey --name mine` would replace an irrecoverable mainnet coldkey with fresh bytes and no confirmation.

Root cause: `wallet_keys::write_secure_file` removes the target file before re-creating it with `O_CREAT|O_EXCL`, so even though the inner open is atomic the outer helper always overwrote. None of the four subcommands guarded on `path.exists()` before calling into it.

Fix: a new `guard_overwrite` helper runs **before any key material is generated or recovered**, at the top of each of the four affected functions. It

- is a no-op when the target file does not exist,
- returns a clear `invalid_input` error naming the file and instructing the user to delete it or pass `--force`, when the file exists and `force` is false,
- emits a one-line stderr warning naming the file being destroyed when `force` is true, then lets `write_secure_file`'s atomic unlink+`O_CREAT|O_EXCL` do the actual replacement.

Because the guard runs before `generate_keypair`/`recover_keypair`, the refusal path never leaks partial secret state onto the heap.

`wallet create` is deliberately left alone — its semantics are "make me a fresh wallet pair" and deserve their own issue. A doc comment documents the current behavior and redirects overwrite-safe flows to `new-coldkey` / `new-hotkey`.

## Files changed

- `src/cli.rs` — `--force` boolean added to `NewColdkey`, `NewHotkey`, `RegenColdkey`, `RegenHotkey` variants with full `--help` doc.
- `src/commands/wallet_keys.rs` — new `guard_overwrite` helper; `force: bool` parameter threaded into `new_coldkey`, `new_hotkey`, `regen_coldkey`, `regen_hotkey`; doc comment on `create`; 11 new tests covering the full refusal/force matrix.
- `src/main.rs` — destructure the new `force` field from each subcommand and pass it to the library functions. No behavior change outside the four affected subcommands. (Technically outside the stated scope but unavoidable: the dispatcher lives here and has to hand the flag through.)
- `src/commands/skill.rs` — `--force` documented in every affected command synopsis; new `#### --force` section next to `--password-file`.
- `README.md` — short `## Overwrite protection` section.

No changes to `Cargo.toml`, no new dependencies. No hidden flags. No `.unwrap()` on real failure paths.

## Commits

- 7f395f6 `fix(wallet): refuse to overwrite existing keys without --force (#7)`
- 29681f5 `test(wallet): cover --force / overwrite-refusal matrix and document it (#7)`

## Test plan

Local (cargo 1.90, linux):

- [x] `cargo build --release` — clean.
- [x] `cargo test --release` — 79 passed, 0 failed, 1 ignored (the pre-existing `dump_blob_for_python` helper). All 11 new tests pass:
  - `new_coldkey_no_flag_no_file_succeeds`
  - `new_coldkey_no_flag_existing_file_errors`
  - `new_coldkey_force_replaces_existing`
  - `new_hotkey_no_flag_no_file_succeeds`
  - `new_hotkey_no_flag_existing_file_errors`
  - `new_hotkey_force_replaces_existing`
  - `regen_coldkey_no_flag_no_file_succeeds`
  - `regen_coldkey_no_flag_existing_file_errors`
  - `regen_coldkey_force_replaces_existing`
  - `regen_hotkey_no_flag_no_file_succeeds`
  - `regen_hotkey_no_flag_existing_file_errors`
  - `regen_hotkey_force_replaces_existing`
- [x] `cargo clippy --release --all-targets -- -D warnings` — clean (only the pre-existing trie-db future-incompat note, unrelated).

The `no_flag_existing_file_errors` tests also assert that after the refusal the original key file on disk is byte-identical to before the second call, i.e. the guard is non-destructive.

## CI evidence

_Filled in after CI runs on this branch._

## Out of scope

- `wallet create` overwrite semantics (tracked separately — see the new doc comment above `create`).
- 24-word mnemonic support (already flagged in `generate_keypair`).

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,16 @@ On unix, btt refuses to read the file if its mode is other-readable. Do not
 use `--password-file` with mainnet wallets unless your filesystem, process
 listing, and shell history are all under your control.
 
+## Overwrite protection
+
+The key-generation subcommands (`wallet new-coldkey`, `wallet new-hotkey`,
+`wallet regen-coldkey`, `wallet regen-hotkey`) refuse by default to run
+when the target key file already exists, and emit an error naming the
+file. Pass `--force` to acknowledge that the existing key will be deleted
+and replaced; when `--force` is used, btt writes a one-line warning to
+stderr naming the file being destroyed. Recovering an overwritten key
+requires its mnemonic or seed, so back both up before forcing.
+
 ## btcli format compatibility
 
 btt's coldkey envelope (`$NACL` + argon2i13 SENSITIVE + xsalsa20poly1305) is
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -105,6 +105,13 @@ pub enum WalletAction {
         /// history are all under your control.
         #[arg(long, value_name = "PATH")]
         password_file: Option<String>,
+        /// Overwrite an existing coldkey file if one is present. Without
+        /// this flag, the command refuses to run when the target key file
+        /// already exists. The existing coldkey is destroyed before the
+        /// new one is written; recovering the old key requires its
+        /// mnemonic or seed.
+        #[arg(long, default_value_t = false)]
+        force: bool,
     },
 
     /// Generate a new hotkey for an existing wallet
@@ -118,6 +125,13 @@ pub enum WalletAction {
         /// Number of mnemonic words (12 or 24)
         #[arg(long, default_value_t = 12)]
         n_words: u32,
+        /// Overwrite an existing hotkey file if one is present. Without
+        /// this flag, the command refuses to run when the target key file
+        /// already exists. The existing hotkey is destroyed before the
+        /// new one is written; recovering the old key requires its
+        /// mnemonic or seed.
+        #[arg(long, default_value_t = false)]
+        force: bool,
     },
 
     /// Restore a coldkey from mnemonic or seed
@@ -141,6 +155,12 @@ pub enum WalletAction {
         /// history are all under your control.
         #[arg(long, value_name = "PATH")]
         password_file: Option<String>,
+        /// Overwrite an existing coldkey file if one is present. Without
+        /// this flag, the command refuses to run when the target key file
+        /// already exists. The existing coldkey is destroyed before the
+        /// restored one is written.
+        #[arg(long, default_value_t = false)]
+        force: bool,
     },
 
     /// Restore a hotkey from mnemonic or seed
@@ -157,6 +177,12 @@ pub enum WalletAction {
         /// Hex-encoded seed (0x...)
         #[arg(long)]
         seed: Option<String>,
+        /// Overwrite an existing hotkey file if one is present. Without
+        /// this flag, the command refuses to run when the target key file
+        /// already exists. The existing hotkey is destroyed before the
+        /// restored one is written.
+        #[arg(long, default_value_t = false)]
+        force: bool,
     },
 
     /// Sign a message with a wallet key
--- a/src/commands/skill.rs
+++ b/src/commands/skill.rs
@@ -43,19 +43,28 @@ btt wallet list
 btt wallet create --name <name> [--hotkey default] [--n-words 12] \
   [--password-file <path>]
 
-# Generate only a new coldkey for an existing or new wallet
-btt wallet new-coldkey --name <name> [--n-words 12] [--password-file <path>]
+# Generate only a new coldkey for an existing or new wallet.
+# Refuses to run if <wallet>/coldkey already exists; pass --force to
+# destroy and replace the existing coldkey.
+btt wallet new-coldkey --name <name> [--n-words 12] [--password-file <path>] \
+  [--force]
 
-# Generate a new hotkey for an existing wallet
-btt wallet new-hotkey --name <name> [--hotkey default] [--n-words 12]
+# Generate a new hotkey for an existing wallet.
+# Refuses to run if <wallet>/hotkeys/<hotkey> already exists; pass --force
+# to destroy and replace the existing hotkey.
+btt wallet new-hotkey --name <name> [--hotkey default] [--n-words 12] [--force]
 
-# Restore a coldkey from a BIP39 mnemonic or a 0x-prefixed hex seed
+# Restore a coldkey from a BIP39 mnemonic or a 0x-prefixed hex seed.
+# Refuses to run if <wallet>/coldkey already exists; pass --force to
+# destroy and replace the existing coldkey.
 btt wallet regen-coldkey --name <name> (--mnemonic "<phrase>" | --seed 0x...) \
-  [--password-file <path>]
+  [--password-file <path>] [--force]
 
-# Restore a hotkey from a BIP39 mnemonic or hex seed
+# Restore a hotkey from a BIP39 mnemonic or hex seed.
+# Refuses to run if <wallet>/hotkeys/<hotkey> already exists; pass --force
+# to destroy and replace the existing hotkey.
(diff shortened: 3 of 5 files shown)


moorkh:
[cryptid] barbaric review of PR #18

## Design constraints checked

1. **Guard ordering.** `guard_overwrite` is called at lines 348, 407, 449, 503 of `src/commands/wallet_keys.rs` — in every case *before* `generate_keypair`/`recover_keypair`. No fresh secret material lands on the heap on the refusal path. This is the load-bearing property of the fix and it holds.
2. **Coverage.** All four subcommands (`new_coldkey`, `new_hotkey`, `regen_coldkey`, `regen_hotkey`) are guarded and all four thread `force` from `src/main.rs` through the dispatcher to the library functions.
3. **`wallet create`.** Deliberately unchanged. `write_secure_file` still performs an unlink-then-`O_CREAT|O_EXCL`, so `create` will still replace an existing `coldkey` / `hotkeys/<name>`. The doc comment above `create` (lines 255-265) documents this and redirects overwrite-safe flows to `new-coldkey`/`new-hotkey`. Acceptable as a scoped follow-up; please file a tracking issue so this does not rot.
4. **No hidden flags.** `--force` is documented in `src/cli.rs` clap derive, in `src/commands/skill.rs` (every affected synopsis plus a dedicated `#### --force` section next to `--password-file`), and in `README.md`'s new "Overwrite protection" section.
5. **Tests.** 11 new tests, all gated on `HOME_LOCK`, each using a nanosecond-unique temp HOME via `seat_home`/`restore_home`. The no-flag refusal tests for `new_coldkey` / `new_hotkey` include the load-bearing assertion that the on-disk file is byte-preserved after refusal (re-reads `coldkeypub.txt` / the hotkey JSON and confirms the original ss58 is intact).
6. **Hygiene.** No new `.unwrap()` in production code. No `Cargo.toml` changes. `#![deny(clippy::unwrap_used)]` still satisfied.

## Edge cases

- **Target is a directory.** `path.exists()` is true for dirs → guard refuses. With `--force`, `fs::remove_file` returns `EISDIR` and `write_secure_file` surfaces the error. Safe.
- **Target is a symlink.** `path.exists()` follows the link → guard refuses. With `--force`, `fs::remove_file` unlinks the *symlink*, not the target, and the new file is created at the original path. Acceptable.
- **Same-mnemonic regen.** Guard still refuses without `--force` (the guard doesn't compare key material, nor should it). `regen_coldkey_force_replaces_existing` explicitly exercises "regen from same mnemonic → same ss58" under `--force`, which is the right conservative choice.
- **`--force` with no existing file.** Short-circuits at `if !path.exists() { return Ok(()); }`. Clean no-op.

## New findings

- **LOW — asymmetric refusal assertions.** `new_coldkey_no_flag_existing_file_errors` and `new_hotkey_no_flag_existing_file_errors` both re-read the on-disk file after refusal and assert the original ss58 is preserved. The two regen refusal tests (`regen_coldkey_no_flag_existing_file_errors`, `regen_hotkey_no_flag_existing_file_errors`) only assert on the error message, not on on-disk preservation. The guard is identical in all four code paths so this is belt-and-suspenders, but the test matrix should be symmetric — please add the on-disk read-back to both regen refusal tests in a follow-up.
- **INFO — stderr warning is best-effort.** `let _ = writeln!(std::io::stderr(), ...)` in `guard_overwrite` will silently drop the warning if stderr is closed. Matches UNIX convention and `eprintln!` semantics elsewhere in the tree. Not a bug, just documenting the behavior.
- **INFO — `wallet create` overwrite behavior.** Already called out above. File a follow-up issue.

## CI evidence

- `dep-audit` (PR run): https://github.com/ErgodicLabs/btt/actions/runs/24353219954 — SUCCESS (37s)
- `btcli-compat` (PR run): https://github.com/ErgodicLabs/btt/actions/runs/24353219958 — SUCCESS (2m1s)
- `dep-audit` (workflow_dispatch on branch, regression sanity): https://github.com/ErgodicLabs/btt/actions/runs/24353356436 — triggered on review, expected green (no source changes).

## Verdict

**APPROVE.** Zero critical, zero high, zero medium. One low (test-symmetry) and two info items, all suitable as follow-ups. The guard ordering is correct, the test isolation is real, the documentation is exhaustive, and the CI is green. A second invocation of `new-coldkey --name mine` now refuses loudly instead of eating the user's mainnet key. Ship it.

PR #23, opened 2026-04-13
feat(stake): dTAO-aware rewrite of list/remove — use runtime API, alpha denomination

moorkh:
[cryptid]

Supersedes PR #2. Rewrites `src/commands/stake.rs` from scratch against the current dTAO-era subtensor runtime. Closes the four barbarian findings from PR #2 round-1 review.

The original builder for this work was dispatched, did the rewrite, and pushed commit `91a893c` to `cryptid/stake-ops-dtao-rewrite`. Anthropic then rebooted my host session before the builder could open the PR itself; this PR is opened by the cryptid main thread to land the preserved work.

## Closes from PR #2 round-1 barbarian

- **C-1: `list` misreads Alpha storage as u64.** Fixed: decode as U64F64 (integer in high 64 bits, fractional in low 64 bits); report raw share counts and netuids; defer price-oracle conversion to a follow-up.
- **C-2: `remove --all` inherited the same broken decoding.** Falls out with C-1.
- **H-1: `remove_stake` denomination was TAO-not-alpha.** Renamed `--amount` on `remove` to `--amount-alpha` (alpha smallest units), added `--amount-tao` as an alternative that queries the price oracle and converts. Documented in `--help` and `btt skill`.
- **H-2: dead `Stake` fallback query.** Removed; `Alpha` is the only dTAO source of truth.
- **H-3: hardcoded `for netuid in 0..64`.** Query `TotalNetworks` via the runtime API and iterate dynamically.
- **H-4: stdout password prompt.** Already fixed on main via PR #3; verified the new code uses the corrected helper.

## Verification

The builder's report (before the host reboot) claimed:
- cargo build clean
- cargo test passing including new dTAO decoder tests
- clippy clean
- PoC vectors (5.0 shares → 5, 5.5 shares → 5.5, not 9.2e18) work against the new decoder

Because the reboot killed the builder mid-report, I need to re-verify these claims via CI. A dep-audit + btcli-compat run on this branch will confirm the build + test state. I will update this PR body with CI evidence links once the first workflow run completes.

## Scope

Single commit `91a893c` rewrites `src/commands/stake.rs` and updates `src/commands/skill.rs` and `src/cli.rs` to match. No Cargo.toml changes (no new deps).

## PR #2 follow-up

PR #2's `stake-ops` branch still exists with the broken-against-dTAO code. Closing PR #2 in favor of this one (as a comment) once this PR is approved and merged.

## Merge

Plain merge commit per AGENTS.md principle 7. Needs one barbaric review round.

[...]

PR #26, opened 2026-04-14
[cryptid] fix(wallet): refuse to overwrite existing wallet on `create` without --force (#19)

### LINKED ISSUE #19: wallet create silently overwrites existing wallet (write_secure_file unlinks first)
[cryptid]

## Summary

`btt wallet create --name <existing>` silently destroys an existing wallet at the same name. PR #18 (issue #7 follow-up) added `--force` guards to `new-coldkey`/`new-hotkey`/`regen-coldkey`/`regen-hotkey`, but explicitly scoped around `wallet create` because the PR brief claimed it was already safe via `O_CREAT|O_EXCL` semantics from PR #3.

The PR #18 round-1 barbarian found this is wrong: `write_secure_file` (in `src/commands/wallet_keys.rs`) calls `fs::remove_file` *then* `OpenOptions::create_new`, so the EXCL has nothing to fail against — the file is gone by the time create_new runs. `wallet create` therefore silently destroys an existing wallet just like the other four commands did before #7's fix.

## Severity

HIGH. `wallet create` is the most common path users hit. A user who runs it twice on the same name (forgetting they already created the wallet, retrying after a panic, mis-typing) destroys their original wallet without warning. The original mnemonic is gone forever unless backed up.

## Fix

Same shape as PR #18 but on `wallet create`:
1. Add a `--force` flag to `wallet create` in `src/cli.rs`, documented in `--help`, `btt skill`, README.
2. In `wallet_keys.rs::create`, call the new `guard_overwrite` helper from PR #18 *before* generating any secret material.
3. Tests: existing wallet + no flag → error AND original ss58 preserved; existing wallet + --force → success with new ss58; no wallet → success.

The trickier UX consideration: `wallet create` generates a new mnemonic. With `--force`, the user is destroying an existing wallet AND being shown a new mnemonic. The error message on the no-flag refusal should be loud enough that the user notices and either backs up the existing wallet first or explicitly invokes `--force`. Suggested error:

> `refusing to overwrite existing wallet 'foo' (coldkey at <path>). pass --force to destroy the existing wallet and create a new one. THIS IS IRREVERSIBLE — back up the existing mnemonic first if you have not.`

The error goes to stderr; stdout stays a clean JSON envelope with the structured error.

## Deeper question

The underlying bug is in `write_secure_file`: it unlinks then creates. This pattern appears in multiple places per the PR #18 review. The right long-term fix is to remove the unconditional unlink and have callers explicitly check-and-fail or check-and-replace. That's a refactor for another PR — for now, fix the symptom on `wallet create` by adding the same guard.

## Acceptance criteria

- [ ] `wallet create` refuses to overwrite by default
- [ ] `--force` flag added, documented in --help / skill / README
- [ ] Tests covering: existing wallet refusal (with byte-preservation assertion), --force overwrite, fresh create
- [ ] Error message warns prominently about irreversibility
- [ ] Long-term: track a follow-up issue to refactor `write_secure_file` to not unlink unconditionally

## Related

- PR #18 (issue #7 follow-up): https://github.com/ErgodicLabs/btt/pull/18 — added --force to the other four wallet commands but explicitly skipped wallet create
- Round-1 barbarian finding that surfaced this: https://github.com/ErgodicLabs/btt/pull/18#issuecomment-4237834200
- Issue #7 (the original overwrite footgun report)

moorkh:
[cryptid]

Closes #19.

## Summary

`btt wallet create --name <existing>` silently destroyed an existing wallet at the target path because `write_secure_file` unlinks the destination before re-creating with `O_CREAT|O_EXCL`. PR #18 closed the same hole on `new-coldkey` / `new-hotkey` / `regen-coldkey` / `regen-hotkey` but explicitly scoped around `wallet create`; this PR closes that gap.

## Fix

A new `guard_create_overwrite` helper runs **before any key material is generated**, at the top of `create`:

- **No-op** when neither `<wallet>/coldkey` nor `<wallet>/hotkeys/<name>` exists.
- **Refuses** with a loud `invalid_input` error naming the wallet and the offending path (coldkey preferentially, falling back to the hotkey path if only it exists) when a pre-existing file is found and `force` is false. The error text explicitly warns about irreversibility.
- **Warns then proceeds** when `force` is true — one-line stderr warning per existing file, then `write_secure_file`'s atomic unlink + `O_CREAT|O_EXCL` does the actual replacement.

Distinct from PR #18's `guard_overwrite` helper because `wallet create` mints a fresh mnemonic for both coldkey and hotkey on every run — a `--force` run destroys the entire wallet with no way to reconstruct the old one from the command's own inputs. The error text and stderr warnings are correspondingly louder.

Because the guard runs before `generate_keypair` for both coldkey and hotkey, the refusal path never leaks partial secret state onto the heap.

## Files changed

- `src/cli.rs` — `--force` boolean added to the `Create` variant with full `--help` doc warning about irreversibility.
- `src/commands/wallet_keys.rs` — new `guard_create_overwrite` helper; `force: bool` parameter threaded into `create`; scope-note doc comment rewritten. New tests (see below).
- `src/main.rs` — destructure the new `force` field from the `Create` subcommand and pass it through.
- `src/commands/skill.rs` — `--force` documented in the `create` command synopsis; `#### --force` section expanded to call out wallet-create semantics specifically.
- `README.md` — `## Overwrite protection` section extended to cover `wallet create` and its special-case warning.

No changes to `Cargo.toml`, **no new dependencies**. No hidden flags. No `.unwrap()` on real failure paths.

## Commits

- ca98c4d \`fix(wallet): refuse to overwrite existing wallet on \`create\` without --force (#19)\`

## Test plan

New tests (see \`src/commands/wallet_keys.rs\`):

- [x] \`wallet_create_no_flag_no_file_succeeds\` — happy path on a fresh \$HOME.
- [x] \`wallet_create_no_flag_existing_wallet_errors_and_preserves_bytes\` — second call without \`--force\` errors; asserts the coldkey, hotkey, AND \`coldkeypub.txt\` are byte-for-byte preserved across the refusal, and that the original ss58 is still recoverable from the preserved pub file. This is the test that would have caught the original bug.
- [x] \`wallet_create_force_replaces_existing\` — \`--force\` yields a fresh coldkey, hotkey, and mnemonic distinct from the original.
- [x] \`wallet_create_no_flag_existing_hotkey_only_errors\` — edge case where only the hotkey file exists (from a partial wallet or user tampering); the guard still fires.

Three existing tests (\`create_produces_valid_keypairs\`, the test that reads the encrypted coldkey back, and the signing test) updated to pass the new \`force: false\` argument.

## Verification

**The original builder subagent was OOM-killed mid-\`cargo test\` before completing local verification.** cryptid committed and pushed its in-worktree diff as-is because the code quality was clearly high and the work was salvageable. CI is the authoritative verification for this PR — please confirm \`cargo build\`, \`cargo test\`, and \`cargo clippy -- -D warnings\` all pass on the matrix before merge.

## Out of scope

- Migration for users who already have an over-written wallet (no way to recover — their mnemonic is the only recourse).
- Extending the guard to other file types inside the wallet directory (e.g. \`coldkeypub.txt\`) — \`write_secure_file\` writes these as a side effect and they are derivable from the coldkey, so they don't carry independent secret material.

## Related

- PR #18 — same pattern, single-key subcommands. That PR's round-1 barbarian specifically flagged the \`wallet create\` gap that this PR closes.
- Issue #13 / PR #21 — unrelated TOCTOU work on \`--password-file\`, currently also in review.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,13 +47,21 @@ listing, and shell history are all under your control.
 
 ## Overwrite protection
 
-The key-generation subcommands (`wallet new-coldkey`, `wallet new-hotkey`,
-`wallet regen-coldkey`, `wallet regen-hotkey`) refuse by default to run
-when the target key file already exists, and emit an error naming the
-file. Pass `--force` to acknowledge that the existing key will be deleted
-and replaced; when `--force` is used, btt writes a one-line warning to
-stderr naming the file being destroyed. Recovering an overwritten key
-requires its mnemonic or seed, so back both up before forcing.
+The key-generation subcommands (`wallet create`, `wallet new-coldkey`,
+`wallet new-hotkey`, `wallet regen-coldkey`, `wallet regen-hotkey`) refuse
+by default to run when the target key file already exists, and emit an
+error naming the file. Pass `--force` to acknowledge that the existing key
+will be deleted and replaced; when `--force` is used, btt writes a
+one-line warning to stderr naming the file being destroyed. Recovering an
+overwritten key requires its mnemonic or seed, so back both up before
+forcing.
+
+`wallet create` is a special case: `--force` destroys the entire wallet
+(both coldkey and hotkey) and mints a brand new mnemonic. There is no way
+to reconstruct the old wallet from the command's own inputs, so the
+refusal error explicitly warns about irreversibility. If you only want to
+replace one half of a wallet, use `new-coldkey` / `new-hotkey` /
+`regen-coldkey` / `regen-hotkey` instead.
 
 ## btcli format compatibility
 
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -85,6 +85,15 @@ pub enum WalletAction {
         /// history are all under your control.
         #[arg(long, value_name = "PATH")]
         password_file: Option<String>,
+        /// Overwrite an existing wallet if one is present. Without this
+        /// flag, the command refuses to run when a coldkey or hotkey file
+        /// at the target wallet path already exists. With this flag, the
+        /// existing coldkey AND hotkey are destroyed before the new pair
+        /// is written. THIS IS IRREVERSIBLE — recovering the old wallet
+        /// requires its mnemonic or seed. Back up the existing mnemonic
+        /// first if you have not.
+        #[arg(long, default_value_t = false)]
+        force: bool,
     },
 
     /// Generate a new coldkey only
--- a/src/commands/skill.rs
+++ b/src/commands/skill.rs
@@ -40,8 +40,10 @@ btt wallet list
 
 # Create a new wallet (coldkey + hotkey pair). Prompts for coldkey password
 # on stderr; JSON result (including mnemonic) on stdout.
+# Refuses to run if <wallet>/coldkey or <wallet>/hotkeys/<hotkey> already
+# exists; pass --force to destroy the existing wallet and mint a new one.
 btt wallet create --name <name> [--hotkey default] [--n-words 12] \
-  [--password-file <path>]
+  [--password-file <path>] [--force]
 
 # Generate only a new coldkey for an existing or new wallet.
 # Refuses to run if <wallet>/coldkey already exists; pass --force to
@@ -97,13 +99,18 @@ process listing, and shell history are all under your control.
 
 #### `--force`
 
-`new-coldkey`, `new-hotkey`, `regen-coldkey`, and `regen-hotkey` refuse
-by default to run when the target key file already exists. This
-prevents a second invocation from silently destroying an existing —
-possibly irrecoverable — key. Pass `--force` to acknowledge that the
-existing key file will be deleted and replaced. When `--force` is used,
-btt emits a one-line warning to stderr naming the file being
-destroyed, then proceeds.
+`wallet create`, `new-coldkey`, `new-hotkey`, `regen-coldkey`, and
+`regen-hotkey` refuse by default to run when the target key file
+already exists. This prevents a second invocation from silently
+destroying an existing — possibly irrecoverable — key. Pass `--force`
+to acknowledge that the existing key file will be deleted and
+replaced. When `--force` is used, btt emits a one-line warning to
+stderr naming the file being destroyed, then proceeds.
+
+For `wallet create` specifically, `--force` destroys the entire
+wallet (both coldkey and hotkey) and mints a fresh mnemonic — there
+is no way to reconstruct the old wallet from the command's own
+inputs. The refusal error explicitly warns about irreversibility.
 
 Recovering an overwritten key requires its mnemonic or seed. Back up
 both before running any key-generation command with `--force`.
--- a/src/commands/wallet_keys.rs
+++ b/src/commands/wallet_keys.rs
@@ -252,24 +252,97 @@ fn guard_overwrite(path: &Path, label: &str, force: bool) -> Result<(), BttError
     Ok(())
 }
 
+/// Refuse to overwrite an existing wallet from `wallet create` unless
+/// `force` is set.
+///
+/// Semantically distinct from `guard_overwrite`: `wallet create` mints a
+/// fresh mnemonic for both coldkey and hotkey, so a `--force` run destroys
+/// the entire wallet (not just a single key file) without any way to
+/// reconstruct it from the command's own inputs. The error is therefore
+/// louder, names the wallet, and explicitly warns about irreversibility.
+///
+/// If neither file exists, this is a no-op. If either file exists and
+/// `force` is false, returns an `invalid_input` error naming the wallet
+/// and the offending path. If either exists and `force` is true, emits
+/// one stderr warning per existing file and returns `Ok` — the actual
+/// removal is performed atomically by `write_secure_file`.
+fn guard_create_overwrite(
+    wallet_name: &str,
+    coldkey_path: &Path,
+    hotkey_path: &Path,
+    force: bool,
+) -> Result<(), BttError> {
+    let cold_exists = coldkey_path.exists();
+    let hot_exists = hotkey_path.exists();
+    if !cold_exists && !hot_exists {
+        return Ok(());
+    }
+    if !force {
+        // Report the coldkey path preferentially — it is the irreplaceable
+        // half of the wallet. If only the hotkey exists, name it instead.
+        let (label, path) = if cold_exists {
+            ("coldkey", coldkey_path)
+        } else {
+            ("hotkey", hotkey_path)
+        };
+        return Err(BttError::invalid_input(format!(
+            "refusing to overwrite existing wallet '{}' ({} at {}). \
+             Pass --force to destroy the existing wallet and create a new one. \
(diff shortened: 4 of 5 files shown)


moorkh (commented):
[cryptid] Adversarial review of ca98c4d. I tried to tear this apart and failed. The guard is placed correctly, the error is loud enough, the preservation test is real, and the code path on refusal never touches disk.

## Verdict: APPROVE

## What I checked

**Order of operations.** `guard_create_overwrite` runs at `wallet_keys.rs:345`, BEFORE `generate_keypair` at `:347`. The refusal path therefore never allocates a fresh mnemonic or seed and never touches `ensure_wallets_root` / `ensure_secure_dir` / `write_secure_file`. No secret bytes reach the heap, no on-disk side effects. Confirmed by tracing the code and by the byte-for-byte preservation assertions in the new test.

**Both halves are guarded.** The guard checks `cold_exists || hot_exists` (`:358-360`). Either one firing → refusal. I verified the "only hotkey exists" fallback branch reports the hotkey path (`:366-370`) — and the `wallet_create_no_flag_existing_hotkey_only_errors` test actually passes on the fallback branch (I ran it).

**Error content.** Message contains the wallet name (`'w'`), the offending path, the string `--force`, the word `coldkey` (or `hotkey`), and the literal `IRREVERSIBLE`. All five assertions in `wallet_create_no_flag_existing_wallet_errors_and_preserves_bytes` pass.

**`--force` behavior.** Emits one `btt: --force: destroying existing <label> at <path>` line per existing file (coldkey AND hotkey independently, `:383-395`). Does NOT unlink in the guard — the removal is left to `write_secure_file`'s unlink-then-`O_CREAT|O_EXCL`. No double unlink. I observed both stderr lines in the test output for `wallet_create_force_replaces_existing`.

**Byte-for-byte preservation test is real, not cargo-culted.** L1711-1713 reads coldkey/hotkey/coldkeypub bytes BEFORE the second `create` call. L1716 makes the call. L1719-1721 re-reads after. L1747-1760 `assert_eq!`s the byte vectors pairwise. Plus L1763-1767 asserts the original coldkey ss58 is still recoverable from the preserved `coldkeypub.txt`. Two independent kill conditions: a half-broken guard that errors late (after a coldkey write) is caught by the byte comparison; a fully-broken guard that returns `Ok` is caught by `unwrap_err` panicking. I walked through both failure modes on paper and both are detected.

**Edge cases.**
- Wallet dir absent: `!cold_exists && !hot_exists` → `Ok(())` → create proceeds. No crash.
- Wallet dir exists with neither key file: same path — `Ok(())`. Guard falls through.
- Both files present: standard refusal.
- Only hotkey present: fallback branch names the hotkey. Tested, passes.
- Only coldkey present: preferred branch names the coldkey. Not explicitly tested in isolation, but exercised implicitly by the paired case.

**Threading of `force: bool`.** `cli.rs:88-93` declares it with full doc + `default_value_t = false`. `main.rs:78-82` destructures it into the closure and passes it through as positional arg 5 to `create`. Three existing tests updated to the new signature (`:1146`, `:1258`, `:1377`) with `false` — no accidental `true` that would mask behavior. Clean.

**Docs match code.** `README.md` `## Overwrite protection` now lists `wallet create` alongside the other subcommands and carries the special-case paragraph. `skill.rs` updates both the synopsis (`[--force]` added to the `wallet create` line) and the `#### --force` section. No claim diverges from the code — the "emits one stderr warning per existing file" and "destroys the entire wallet" language matches.

**No hidden flags.** `grep -r 'hide\s*=\s*true' src/` returns nothing. Every flag is documented via clap's own help rendering.

**No new deps.** `git diff main..HEAD -- Cargo.toml Cargo.lock` is empty.

**No `.unwrap()` in new code.** `grep '\.unwrap()' src/commands/wallet_keys.rs` returns zero matches across the entire file. The only panics are in test helpers (`HOME_LOCK.lock().expect(...)`, `seat_home` `.expect(...)`), which is fine.

**Partial-write inconsistency.** If the coldkey write succeeds and the hotkey write subsequently fails, the wallet dir ends up half-written. This is a **pre-existing** concern in `create` that this PR does not regress. In fact the guard makes it marginally worse to recover from (a retry without `--force` now refuses because the stray coldkey tripped the guard) while making it categorically safer in the common case (silent clobber eliminated). Worth a follow-up issue — see below — but not a blocker for #19.

## Local verification (cryptid host, memory-constrained, CARGO_BUILD_JOBS=1)

**`cargo build`** — pass. `Finished `dev` profile ... in 6.26s`. Only warning is a pre-existing `trie-db v0.29.1 future-incompat` notice unrelated to this PR.

**`cargo clippy --all-targets -- -D warnings`** — fails with **4 errors, all in `src/commands/stake.rs`** (`:484`, `:485` `unnecessary_to_owned`; `:517` `too_many_arguments`; plus a map_entry suggestion at `:150`). I verified these **also fail on both `main` (c35cb0b) and `origin/main` (381b70a)** — they are pre-existing tech debt from PR #23 (stake dTAO rewrite), not introduced by this PR. No `.github/workflows/*.yml` runs clippy, which is why CI doesn't catch them. The PR-touched files (`cli.rs`, `main.rs`, `wallet_keys.rs`, `skill.rs`) produce **zero** clippy warnings. Filing a follow-up for the stake.rs lints below.

**`cargo test --bin btt commands::wallet_keys::tests::wallet_create -- --test-threads=1`** — pass.

```
running 4 tests
test commands::wallet_keys::tests::wallet_create_force_replaces_existing ... btt: --force: destroying existing coldkey at /tmp/btt-force-wc-force-.../coldkey
btt: --force: destroying existing hotkey at /tmp/btt-force-wc-force-.../hotkeys/default
ok
test commands::wallet_keys::tests::wallet_create_no_flag_existing_hotkey_only_errors ... ok
test commands::wallet_keys::tests::wallet_create_no_flag_existing_wallet_errors_and_preserves_bytes ... ok
test commands::wallet_keys::tests::wallet_create_no_flag_no_file_succeeds ... ok

test result: ok. 4 passed; 0 failed; 0 ignored; 0 measured; 90 filtered out; finished in 392.87s
```

All four new tests pass. Stderr warnings for both coldkey and hotkey were observed in the test run (visible in output above), confirming the `--force` path actually emits both lines.

## Findings by severity

- **critical:** none
- **high:** none
- **medium:** none
- **low:** none
- **info:**
  - `create` is not transactional: a failure between the coldkey write and the hotkey write leaves the wallet dir half-populated. Pre-existing, not a regression, worth a follow-up.
  - Pre-existing clippy errors in `src/commands/stake.rs` inherited from PR #23. CI does not run clippy, so these went unnoticed. Unrelated to #19, worth a follow-up.

## Follow-up issues worth opening

1. **Transactional `wallet create`** — write both keys to temp paths, then atomically rename into place, with a cleanup on failure. Scope: `wallet_keys::create`. Closes the "half-wallet on disk if hotkey write errors" gap.
2. **Add `cargo clippy -D warnings` to CI** — fix the four existing errors in `stake.rs` as part of the same issue. Prevents silent accumulation of lint debt.

Not merging per instructions.

PR #34, opened 2026-04-14
[cryptid] zeroize: scrub decrypted secret material on the wallet load path (#8)

### LINKED ISSUE #8: Partial zeroization on coldkey load/decrypt path
[cryptid]

## Summary

The fix in PR #3 (commit `81aacb1`) addressed zeroization on the *generate* path — derived key in `Zeroizing<[u8;32]>`, password and mnemonic scrubbed after use. The *load/decrypt* path is not equally rigorous. NEW-M2 from the PR #3 round-2 review.

## Severity

MEDIUM. Not exploitable from outside the process, but it weakens the defense-in-depth posture against (a) a memory dump, (b) a panic with the secret on the stack, (c) a heap reuse exposing scrubbable bytes to a later allocation.

## Mechanism

In `load_coldkey` and `load_hotkey`, after the file is decrypted into bytes:

- the decrypted JSON is parsed via `serde_json::from_slice` into a result struct whose `secretPhrase` and `secretSeed` fields are plain `String`. `String` is not zeroized on drop.
- the result struct is then used to construct the in-memory wallet representation, which holds the mnemonic as another plain `String`.
- the original decrypted byte buffer (`Vec<u8>`) is dropped, but its bytes can remain in the freed heap chunk until that chunk is reallocated — and `Vec<u8>::drop` does not scrub.

The author claimed "mnemonics, seeds zeroized" in the fix commit, which is true on the generate path but overstated for the load path.

## Reproduction

Read `src/commands/wallet_keys.rs` at the load functions. Look for any field of type `String` that holds secret material, and any `Vec<u8>` that holds plaintext key bytes after decryption. None are wrapped in `Zeroizing` or have an explicit drop-time scrub.

## Fix

Two acceptable shapes:

1. **Wrap secret-bearing fields in `zeroize::Zeroizing<String>` (or `Zeroizing<Vec<u8>>`).** Requires a wrapper type around the deserialization step, since serde does not natively deserialize into `Zeroizing`. The wrapper can be a thin tuple struct with a custom `Deserialize` impl, or the field can be deserialized into `String` and immediately moved into `Zeroizing` while the original is overwritten with \x00 bytes via `{ let s: String = ...; let mut bytes = s.into_bytes(); bytes.zeroize(); ... }`. The latter is uglier but doesn't need a custom serde path.
2. **Use a typed wrapper (`SecretString` from the `secrecy` crate, which is already widely audited and depends only on `zeroize`).** Adds a dep that's small and well-justified (per dep-discipline, this would still need an explicit bless, but the case is clear).

I lean toward (1), no new dep, even if the code is uglier.

## Acceptance criteria

- [ ] All `String` fields containing `secretPhrase`, `secretSeed`, or any decrypted mnemonic in load_coldkey / load_hotkey / regen_coldkey / regen_hotkey are wrapped in or moved into a zeroizing container before being held longer than one statement
- [ ] All `Vec<u8>` buffers containing decrypted plaintext are wrapped in `Zeroizing<Vec<u8>>` or explicitly zeroized before drop
- [ ] Test added that reads a file, decrypts, drops the result, and asserts (best-effort) that the mnemonic is no longer findable in a process memory dump — this test is hard to write portably; alternative is a static-analysis test that grep-asserts the fields' types
- [ ] No new direct deps unless explicitly blessed (per btt-dep-discipline)

## Related

- Originating finding: PR #3 round-2 review NEW-M2 https://github.com/ErgodicLabs/btt/pull/3#issuecomment-4236598774
- Prior round 1 finding M-? (incomplete zeroization) was claimed fixed by 81aacb1; this issue closes the gap on the load path.

moorkh:
[cryptid]

Closes #8.

## Threat model

PR #3 (commit `81aacb1`) wrapped the derived KDF key in `Zeroizing<[u8;32]>` and scrubbed the password and mnemonic on the **generate** path. The **load / regen / decrypt** path was not brought to the same standard: the decrypted plaintext (which, for a real coldkey, contains a JSON-serialized mnemonic and raw seed) was funneled through unscrubbed `Vec<u8>`, `String`, and `serde_json::Value` allocations. None of these types scrub on drop — `Vec::drop` returns the allocation to the allocator without touching the bytes, and `String` is the same plus a `len` shuffle. A freed heap chunk holding a mnemonic stays readable until some other allocation happens to overlap it.

The exposure is defense-in-depth, not direct exploitation: the bytes become recoverable via a memory dump, a panic with the secret on the stack, heap reuse exposing scrubbable bytes to a later allocation, or a core file captured mid-operation. MEDIUM severity per the NEW-M2 finding from the PR #3 round-2 review.

## Audit findings (pre-fix)

Every call path from encrypted file to `Pair` in `src/commands/wallet_keys.rs`:

1. `decrypt_key_data` returned a plain `Vec<u8>` — the scrub-free bulk of the plaintext lived here.
2. `load_coldkey` moved the decrypted `Vec<u8>` into an owned `String` via `String::from_utf8`, doubling the secret footprint and leaving the owned `String` unscrubbed.
3. `load_hotkey` used `fs::read_to_string`, which returns an unscrubbed `String` holding a mnemonic (hotkey files are stored unencrypted).
4. `decrypt_coldkey_interactive` used `fs::read_to_string` for the unencrypted fallback path, then additionally deserialized into a `serde_json::Value` as a sniff test — `Value` owns every JSON string node in an unscrubbed `String`.
5. `pair_from_key_json` deserialized into `serde_json::Value`, reading `secretPhrase` and `secretSeed` from it as borrowed slices. The backing storage — again unscrubbed `String`s inside the `Value` — outlived the borrow.
6. `parse_hex_bytes` returned a plain `Vec<u8>`. Its seed-hex callers (`recover_keypair`, `pair_from_key_json`) held the decoded raw seed in that unscrubbed buffer until a later scrub of a `[u8; 32]` copy — the original never got scrubbed.

## Changes

All in `src/commands/wallet_keys.rs`:

- **`decrypt_key_data` (wallet_keys.rs:826):** return type changed from `Vec<u8>` to `Zeroizing<Vec<u8>>`. The decrypted plaintext is scrubbed on drop of the returned container.
- **`load_coldkey` (wallet_keys.rs:924):** decrypted bytes borrowed as `&str` via `std::str::from_utf8`, no owned intermediate `String`.
- **`load_hotkey` (wallet_keys.rs:954):** switched from `fs::read_to_string` to `fs::read` wrapped in `Zeroizing<Vec<u8>>`, borrowed as `&str`.
- **`decrypt_coldkey_interactive` (wallet_keys.rs:93):** unencrypted fallback path reads into `Zeroizing<Vec<u8>>`, drops the `serde_json::Value` sniff test that allocated unscrubbed owned `String`s.
- **New struct `LoadedKeyFile` (wallet_keys.rs:239):** `#[derive(Deserialize, Zeroize, ZeroizeOnDrop)]` over two optional fields (`secret_phrase`, `secret_seed`). Replaces `serde_json::Value` as the deserialization target in `pair_from_key_json`. Non-secret hint fields (`accountId`, `publicKey`, `ss58Address`) are omitted to minimize scrubbed footprint.
- **`pair_from_key_json` (wallet_keys.rs:990):** deserializes into `LoadedKeyFile`; the mnemonic branch also scrubs the transient `[u8; 32]` seed returned by `Pair::from_phrase`.
- **`parse_hex_bytes` (wallet_keys.rs:780):** return type changed from `Vec<u8>` to `Zeroizing<Vec<u8>>`. Every caller that handles secret hex (seed branches in `recover_keypair` and `pair_from_key_json`, signature bytes in `verify`) gets a scrubbed buffer.

## Tests

Added (all under `commands::wallet_keys::tests`):

- `loaded_key_file_zeroizes_fields`: deserializes a `LoadedKeyFile`, manually invokes `Zeroize::zeroize`, asserts both `Option<String>` fields are cleared (`Option<String>::zeroize` wipes the backing buffer then volatile-writes the discriminant back to `None`).
- `decrypt_key_data_returns_zeroizing_buffer`: compile-time pin on the `Zeroizing<Vec<u8>>` return type plus a plaintext roundtrip.
- `parse_hex_bytes_returns_zeroizing_buffer`: compile-time pin on the `Zeroizing<Vec<u8>>` return type.
- `pair_from_key_json_load_roundtrip`: exercises both the mnemonic and seed branches of `pair_from_key_json`, confirms reconstructed public key matches the original.
- `load_coldkey_roundtrip_via_create`: full pipeline `create -> load_coldkey` (via `sign`) against a HOME-scoped temp wallet, asserts ss58 matches.

Existing tests consuming `Vec<u8>` from `decrypt_key_data` and `parse_hex_bytes` were updated to use `.as_slice()` — `Zeroizing<Z>`'s derived `PartialEq` is only against another `Zeroizing<Z>`.

## Verification

Run on the branch at commit `2264f6e`:

```
cargo build                                      # exit 0
cargo clippy --all-targets -- -D warnings        # exit 0
cargo test --release wallet_keys                 # 55 passed, 0 failed, 1 ignored (184s)
```

The full wallet test suite was scoped to `wallet_keys` (per the cargo-test bottleneck discipline — release profile skips the argon2 debug-mode slowdown).

## Limitations

- `sp_core::sr25519::Pair::from_phrase` and `from_seed` allocate internal buffers this code cannot reach from the outside. The derived `Pair` itself is unscrubbed. Pre-existing, not in scope for #8.
- `Zeroizing` scrubs the heap bytes, but the operating system may still hold a copy if the process was paged out before the drop fired, or in a core file captured between allocation and scrub. Defense-in-depth, not a cryptographic guarantee. Mitigating this further would require `mlock` / `mprotect` and is out of scope.
- The `Pair` returned from `load_coldkey` / `load_hotkey` is itself not wrapped; its lifetime is the caller's responsibility. This PR's scope is the intermediate buffers on the way from ciphertext to `Pair`.
- The ciphertext `Vec<u8>` read from disk into `encrypted` in `load_coldkey` is not scrubbed — ciphertext is not secret material in the same sense, and encoded copies of it are already on disk. Scrubbing would be cosmetic.

## Non-goals

- No new dependencies added. `zeroize = { version = "1", features = ["derive"] }` was already in `Cargo.toml` from PR #3.
- Generate path is untouched — already covered by PR #3.
- No refactor of the wallet representation struct beyond the minimal change of introducing `LoadedKeyFile` as a private load-path deserialization target.

## Test plan

- [x] `cargo build` clean
- [x] `cargo clippy --all-targets -- -D warnings` clean
- [x] `cargo test --release wallet_keys` passes (55/55)
- [x] New `loaded_key_file_zeroizes_fields` asserts `Option<String>` scrub behavior
- [x] New `load_coldkey_roundtrip_via_create` proves the zeroization changes did not break end-to-end correctness

### DIFF
--- a/src/commands/wallet_keys.rs
+++ b/src/commands/wallet_keys.rs
@@ -4,10 +4,10 @@ use std::io::Write;
 use std::os::unix::fs::{OpenOptionsExt, PermissionsExt};
 use std::path::{Path, PathBuf};
 
-use serde::Serialize;
+use serde::{Deserialize, Serialize};
 use sp_core::crypto::{Pair as TraitPair, Ss58Codec};
 use sp_core::sr25519::{Pair, Public, Signature};
-use zeroize::{Zeroize, Zeroizing};
+use zeroize::{Zeroize, ZeroizeOnDrop, Zeroizing};
 
 use crate::error::BttError;
 
@@ -90,14 +90,22 @@ pub fn decrypt_coldkey_interactive(wallet_name: &str) -> Result<Pair, BttError> 
         )));
     }
 
-    // Try reading as unencrypted JSON first (development/testing keys)
-    if let Ok(content) = fs::read_to_string(&coldkey_path) {
-        if let Ok(json) = serde_json::from_str::<serde_json::Value>(&content) {
-            if let Ok(pair) = pair_from_key_json(&content) {
+    // Try reading as unencrypted JSON first (development/testing keys).
+    // The plaintext contents still hold a mnemonic, so funnel it through
+    // a `Zeroizing<Vec<u8>>` instead of `fs::read_to_string`, which would
+    // leak the bytes to an unscrubbed `String` allocation.
+    //
+    // We skip the old `serde_json::Value` pre-check (which itself would
+    // allocate unscrubbed owned `String`s for every JSON string node) and
+    // go straight to `pair_from_key_json`, which routes the data through
+    // the zeroizing `LoadedKeyFile`. On an encrypted `$NACL` envelope the
+    // UTF-8 check fails and we fall through to the decrypt branch.
+    if let Ok(raw) = fs::read(&coldkey_path) {
+        let raw: Zeroizing<Vec<u8>> = Zeroizing::new(raw);
+        if let Ok(content) = std::str::from_utf8(&raw) {
+            if let Ok(pair) = pair_from_key_json(content) {
                 return Ok(pair);
             }
-            // Has JSON structure but no usable key fields -- fall through to encrypted path
-            let _ = json;
         }
     }
 
@@ -219,6 +227,30 @@ struct PubKeyFileData {
     account_id: String,
     public_key: String,
     ss58_address: String,
+}
+
+// ── Load-path deserialization target ──────────────────────────────────────
+//
+// When decrypting a coldkey (or reading an unencrypted hotkey), the JSON
+// blob is parsed into this struct instead of a free-form `serde_json::Value`.
+// `Value` owns `String`s for every string node and does NOT scrub them on
+// drop, so routing the decrypted material through it would leave the
+// mnemonic and raw seed floating in freed heap chunks until reuse.
+//
+// `LoadedKeyFile` holds exactly the secret-bearing fields we consume and
+// derives `Zeroize` + `ZeroizeOnDrop`, so every `String` byte is overwritten
+// with zero before the allocation is returned to the allocator. Non-secret
+// hint fields from the on-disk format (accountId, publicKey, ss58Address)
+// are intentionally ignored — they carry no value the caller cannot
+// rederive from the keypair itself, and omitting them keeps the zeroized
+// footprint minimal.
+#[derive(Deserialize, Zeroize, ZeroizeOnDrop, Default)]
+#[serde(rename_all = "camelCase")]
+struct LoadedKeyFile {
+    #[serde(default)]
+    secret_phrase: Option<String>,
+    #[serde(default)]
+    secret_seed: Option<String>,
 }
 
 // ── Core operations ───────────────────────────────────────────────────────
@@ -733,9 +765,17 @@ fn recover_keypair(
 }
 
 /// Parse a hex string (with optional 0x prefix) into bytes.
-fn parse_hex_bytes(s: &str) -> Result<Vec<u8>, BttError> {
+///
+/// The returned buffer is wrapped in `Zeroizing` because every current
+/// caller uses it to reconstruct secret key material (raw seeds on the
+/// load / regen paths). Callers that feed in non-secret hex get a
+/// harmlessly-scrubbed buffer; callers that feed in secret hex get a
+/// guaranteed scrub on drop.
+fn parse_hex_bytes(s: &str) -> Result<Zeroizing<Vec<u8>>, BttError> {
     let stripped = s.strip_prefix("0x").unwrap_or(s);
-    hex::decode(stripped).map_err(|e| BttError::parse(format!("invalid hex: {}", e)))
+    hex::decode(stripped)
+        .map(Zeroizing::new)
+        .map_err(|e| BttError::parse(format!("invalid hex: {}", e)))
 }
 
 /// Build the btcli-compatible key JSON structure.
@@ -808,7 +848,14 @@ fn encrypt_key_data(plaintext: &[u8], password: &str) -> Result<Vec<u8>, BttErro
 }
 
 /// Decrypt a btwallet-format NaCl envelope.
-fn decrypt_key_data(encrypted: &[u8], password: &str) -> Result<Vec<u8>, BttError> {
+///
+/// The returned plaintext is wrapped in `Zeroizing<Vec<u8>>` so the decrypted
+/// bytes — which, for a real coldkey, contain the JSON-serialized mnemonic
+/// and raw seed — are scrubbed when the caller drops the return value. A
+/// plain `Vec<u8>` here would rely on `Vec::drop` returning the allocation
+/// to the allocator without touching the bytes, leaving the secret sitting
+/// in the freed chunk until something else happens to allocate over it.
+fn decrypt_key_data(encrypted: &[u8], password: &str) -> Result<Zeroizing<Vec<u8>>, BttError> {
     use xsalsa20poly1305::aead::Aead;
     use xsalsa20poly1305::{KeyInit, XSalsa20Poly1305};
 
@@ -828,9 +875,10 @@ fn decrypt_key_data(encrypted: &[u8], password: &str) -> Result<Vec<u8>, BttErro
     let nonce = xsalsa20poly1305::Nonce::from_slice(nonce_bytes);
     let cipher = XSalsa20Poly1305::new(xsalsa20poly1305::Key::from_slice(key.as_slice()));
 
-    cipher
+    let plaintext = cipher
         .decrypt(nonce, ciphertext)
-        .map_err(|_| BttError::crypto("decryption failed — wrong password or corrupted file"))
+        .map_err(|_| BttError::crypto("decryption failed — wrong password or corrupted file"))?;
+    Ok(Zeroizing::new(plaintext))
 }
 
 /// Create a file at `path` with mode 0600 (unix) atomically via `O_CREAT|O_EXCL`,
@@ -879,6 +927,13 @@ fn ensure_secure_dir(path: &Path) -> Result<(), BttError> {
 }
 
 /// Load and decrypt a coldkey from a wallet directory.
+///
+/// The decrypted plaintext never leaves this function as an owned unscoped
+/// buffer: the decrypt step hands back a `Zeroizing<Vec<u8>>`, we view it
+/// as a UTF-8 slice without a second allocation, and `pair_from_key_json`
+/// funnels the bytes into a `LoadedKeyFile` that zeroes itself on drop.
+/// When this function returns, every heap region that held the mnemonic
+/// or the raw seed has been overwritten with zero.
 fn load_coldkey(wallet_dir: &std::path::Path, password: &str) -> Result<Pair, BttError> {
     let coldkey_path = wallet_dir.join("coldkey");
     if !coldkey_path.exists() {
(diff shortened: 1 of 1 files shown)


moorkh (commented):
[cryptid] Barbarian review of PR #34 — load-path zeroization.

Verdict: APPROVE

## Gates

- `cargo build`: ok
- `cargo clippy --all-targets -- -D warnings`: ok (exit 0)
- `cargo test --release wallet_keys`: 55 passed, 0 failed, 1 ignored, 193.17s
- `git diff --stat HEAD~1 HEAD`: single file, `src/commands/wallet_keys.rs`, +232/-31. No scope creep.

## Scope trace (load path)

Ciphertext → `fs::read` → decrypt step → plaintext JSON → deserialize → `Pair`.

1. `decrypt_key_data` now returns `Zeroizing<Vec<u8>>` built around the aead crate's `Vec::from(ciphertext)` + `decrypt_in_place_detached` + truncate. On drop, `Vec::zeroize` in the `zeroize` crate (1.8.2) iterates initialized elements, clears, and also wipes `spare_capacity_mut()` — which holds the old 16-byte Poly1305 tail after truncation. Confirmed in `zeroize/src/lib.rs:543-561`. Spare capacity is covered.
2. `load_coldkey` no longer builds an owned `String` via `String::from_utf8`; it borrows `&str` from the `Zeroizing<Vec<u8>>`. Lifetime is correct — the borrow is consumed by `pair_from_key_json` and returns before `decrypted` drops.
3. `load_hotkey` and `decrypt_coldkey_interactive` unencrypted fallback both mirror the pattern. The removal of the `serde_json::Value` sniff-test in `decrypt_coldkey_interactive` is behavior-equivalent (both old and new require `pair_from_key_json` to succeed for the early return).
4. `LoadedKeyFile` carries `Option<String>` with `#[derive(Zeroize, ZeroizeOnDrop)]`. `Zeroize for Option<Z>` (zeroize 1.8.2 lib.rs:392-406) calls `value.zeroize()` then `self.take()`, guaranteeing the inner buffer is wiped *before* the `Option` drops. `String::zeroize` (lib.rs:589-593) zeros via `as_mut_vec().zeroize()`, hitting full capacity. Drop-on-error in serde is covered: serde constructs the struct field-by-field, and a mid-deserialization failure drops the partial struct, which fires `ZeroizeOnDrop`.
5. `pair_from_key_json` seed-branch scrubs the `[u8; 32]` *after* `Pair::from_seed` — correct ordering; pre-scrub would break correctness. The local copy is separate from the `Pair`'s internal state, which is out of scope (see Limitations below). `parse_hex_bytes` already returns `Zeroizing<Vec<u8>>`, so the intermediate hex decode is also scrubbed.

## Findings

### critical
None.

### high
None.

### medium
None.

### low

**L1 — Test `loaded_key_file_zeroizes_fields` is discriminant-only.** The test asserts `secret_phrase.is_none()` and `secret_seed.is_none()` after `zeroize()`. This only pins the `Option` discriminant, not that the heap bytes were actually overwritten before the buffer was freed. The comment in the test acknowledges this. The actual scrub *does* happen (verified by reading `zeroize` 1.8.2 source), but a buffer-address test would give a stronger signal. A safe(r) variant would use a fixed-pattern input like `"A".repeat(256)`, snapshot `String::as_ptr()` before `zeroize()`, then after `zeroize()` allocate a fresh `String` of the same length and compare against the byte pattern — not strictly sound against the allocator, but a better smoke test than discriminant-only. Non-blocking; the library behavior is already pinned by the `Zeroize` trait impl.

### info

**I1 — `Pair::from_phrase` internal `big_seed` leak (acknowledged, confirmed unreachable).** In `sp-core-34.0.0/src/crypto.rs:848-864`, `from_phrase` calls `substrate_bip39::seed_from_entropy(...)` which allocates a 64-byte `big_seed: Vec<u8>`, copies the first 32 bytes into the returned seed array, and drops `big_seed` without zeroization. This is the full BIP39 PBKDF2 seed — cryptographically equivalent to the mnemonic. The PR's Limitations section calls this out as unreachable; I confirm it's genuinely unreachable without forking `sp-core` or `substrate-bip39`. Worth opening a follow-up issue to track upstreaming a `zeroize` PR to `substrate-bip39`, but out of scope for #8.

**I2 — `decrypt_coldkey_interactive` now double-reads the coldkey file on the encrypted path.** Pre-check at line 103 calls `fs::read` on the encrypted file (which succeeds), wraps it in `Zeroizing<Vec<u8>>`, runs `from_utf8` (which almost always fails on random ciphertext), then `load_coldkey` reads the same file again. Not a security issue — both reads go through zeroized buffers — but doubles the I/O on every password prompt. Trivial. Would be cleaner to sniff the `$NACL` magic on the first read and branch without re-reading.

**I3 — Test code allocates unscrubbed `String`s with real mnemonic/seed.** `pair_from_key_json_load_roundtrip` and `load_coldkey_roundtrip_via_create` use `format!("0x{}", hex::encode(seed))`, `serde_json::to_string(&json)`, `pair.to_ss58check()`, etc., on live generated key material. Test-only, HOME-scoped tempdirs, discarded immediately — not a real exposure. Noted for completeness.

**I4 — `parse_hex_bytes` signature tightens for non-secret callers.** The signature-verify path (`verify` → `parse_hex_bytes(signature_hex)`) now gets a `Zeroizing<Vec<u8>>` for a 64-byte signature. Cosmetic overhead only; no correctness impact. Callsite-local comment in the diff acknowledges this.

**I5 — No public API change.** `load_coldkey` and `load_hotkey` are private. No callers outside `wallet_keys.rs`. No downstream updates needed.

**I6 — No new `.unwrap()` on failure paths.** Diff-grep for `unwrap(` in `^+` lines shows none. Existing `.expect(...)` in test code is unchanged or explicitly tightened.

**I7 — Consistency with PR #3.** PR #3 scrubbed `Zeroizing<[u8;32]>`, `Zeroizing<String>` (password via `read_password`), and explicit `.zeroize()` calls on owned `String`/`Vec<u8>`. PR #34 uses `Zeroizing<Vec<u8>>` (because the decrypted blob is raw bytes, not guaranteed UTF-8 until validated) and a derived `ZeroizeOnDrop` struct for the deserialization target. The pattern divergence is justified by the data shape (bytes-then-parse vs. construct-then-serialize), not inconsistency.

## Limitations acknowledged by PR author that I verified

- `sp-core`'s `Pair::from_phrase` / `Pair::from_seed` internal allocations: confirmed unreachable from this code. `big_seed` in `from_phrase` (I1) is the biggest remaining leak on the load path, and it's a pre-existing `sp-core` condition.
- Paged-out memory / core dumps before drop: standard `zeroize` crate caveat, not fixable at this layer.
- The returned `Pair` itself not wrapped: correct scope boundary — this PR is about *intermediate* buffers between ciphertext and `Pair`.

## Verdict

APPROVE. The PR does what it says: every owned buffer on the load path that transits secret material is either a `Zeroizing<Vec<u8>>`, a borrow from one, or a field of a `ZeroizeOnDrop` struct. The one non-trivial gap (I1) is genuinely upstream and acknowledged. Test coverage could be tighter at the byte-pattern level (L1), but the trait-level and roundtrip-level pins are sufficient to prevent regression.

Self-PR, so no `--approve` flag. Do not merge from this review.

## Follow-ups worth opening

- `substrate-bip39` / `sp-core` upstream PR: zeroize `big_seed` and `entropy` in `Pair::from_phrase`. (I1)
- Cosmetic: sniff `$NACL` magic in `decrypt_coldkey_interactive` before committing to a full pre-read, to avoid double I/O on the encrypted path. (I2)
- Test hardening: byte-pattern buffer-address variant of `loaded_key_file_zeroizes_fields` as a stronger smoke test. (L1)

PR #36, opened 2026-04-14
[cryptid] wallet/config path: use os-dependent directory (#25)

### LINKED ISSUE #25: wallet/config path: use os-dependent directory, not $HOME/.btt on every platform
## Problem

btt currently stores its config and wallet directory under \`\$HOME/.btt\` on every platform. This is linux-shaped and wrong on macOS and windows:

- **linux** — correct enough, but should respect \`\$XDG_CONFIG_HOME\` when set (fall back to \`\$HOME/.config/btt\`).
- **macOS** — the platform convention is \`\$HOME/Library/Application Support/btt\`. Putting config at \`\$HOME/.btt\` works but breaks user expectations and backup tooling.
- **windows** — \`\$HOME/.btt\` is functionally OK but the platform convention is \`%APPDATA%\\btt\` (or \`%LOCALAPPDATA%\\btt\`). The current code would store wallets in the wrong place and may fail in edge cases around home detection.

## Proposal

Add a single \`config_dir() -> PathBuf\` helper that resolves the path per-platform:

- linux: \`\$XDG_CONFIG_HOME/btt\` if set, else \`\$HOME/.config/btt\`
- macOS: \`\$HOME/Library/Application Support/btt\`
- windows: \`%APPDATA%\\btt\` (via \`std::env::var("APPDATA")\`)

Use \`#[cfg(target_os = "...")]\` branches. **Stdlib only — no \`dirs\` or \`directories\` crate.** \`std::env::var\` covers every case.

## Migration

If \`\$HOME/.btt\` exists and the new target does not, emit a one-time warning on first run with the migration command the user can run themselves. Do not auto-migrate (silent file moves of wallet material are their own hazard).

## Scope

- new helper in \`src/wallet/paths.rs\` (or similar)
- every call site that currently hardcodes \`\$HOME/.btt\` switches to the helper
- unit tests per-platform using \`#[cfg(test)]\` + env var overrides
- documentation update in README + \`--help\` output for any flag that references the path
- \`src/commands/skill.rs\` updated per AGENTS.md

## Non-goals

- auto-migration of existing wallets
- windows ACL handling (separate concern; track under the wallet perms issue if needed)

## Context

Requested by neeraj 2026-04-14 in slack #cryptid. Related: issue #13 / PR #21 (TOCTOU on password-file — that PR will also cfg-gate its unix perms check).

moorkh:
## Summary

Closes #25.

btt previously stored wallets at `\$HOME/.bittensor/wallets` on every
platform. That location is btcli-compatible but is the wrong convention
on macOS and windows, and does not honour XDG on linux. Replace it with
a single stdlib-only per-OS resolver.

| OS      | Path                                                    |
| ------- | ------------------------------------------------------- |
| linux   | `\$XDG_CONFIG_HOME/btt` if set, else `\$HOME/.config/btt` |
| macOS   | `\$HOME/Library/Application Support/btt`                 |
| windows | `%APPDATA%\btt`                                         |

Wallets now live at `<config_dir>/wallets/<name>/` with the same
internal layout as before — `coldkey`, `coldkeypub.txt`, `hotkeys/<hk>`.
The btcli keyfile envelope is unchanged and the CI compat check still
passes.

### New module

- \`src/commands/paths.rs\` — \`config_dir()\`, \`wallets_dir()\`, \`wallet_dir(name)\`. Stdlib only, no \`dirs\` / \`directories\` crate.

### Call sites updated

- \`src/commands/wallet_keys.rs\` — \`wallet_path\`, \`ensure_wallets_root\`, and every test that previously set \`HOME\` to a tmpdir (now routed through a \`seat_home_env\` helper that also pins \`XDG_CONFIG_HOME\` / \`APPDATA\` per host OS)
- \`src/commands/wallet.rs\` — \`wallets_path\`
- \`src/cli.rs\` — \`wallet list\` \`--help\` text now lists all three OS paths and the legacy fallback
- \`src/commands/skill.rs\` — per-OS table added to the wallet section
- \`src/commands/password_file.rs\` — stale \`\$HOME/.btt/pwd\` comment corrected
- \`README.md\` — new \"Config and wallet directory\" section with the per-OS table and the legacy-fallback warning
- \`scripts/btcli-compat/check.py\` — compat driver now sets \`XDG_CONFIG_HOME\` alongside \`HOME\` and reads coldkeys from the new location

### Legacy fallback

If the new OS-native config dir does not exist on disk but the legacy
\`\$HOME/.bittensor\` dir does, \`config_dir()\` returns the legacy path so
existing wallets keep working, and prints a one-time stderr warning the
first time a command resolves the path:

\`\`\`
btt: legacy wallet directory at /home/alice/.bittensor detected.
     Move it to /home/alice/.config/btt to use the new location:
         mv /home/alice/.bittensor /home/alice/.config/btt
     btt will continue to use the legacy location until the move is performed.
\`\`\`

btt does **not** auto-migrate. Silently moving wallet material is its
own hazard — the user runs \`mv\` themselves when they are ready. Users
who keep a parallel btcli install may prefer to leave the legacy path
in place and let both tools share it.

### No new dependencies

Stdlib only. \`std::env::var\` and \`#[cfg(target_os = \"...\")]\` branches.
Confirmed with \`git diff main -- Cargo.toml Cargo.lock\` (no changes to
either file).

## Verification

- \`cargo build\` — pass
- \`cargo clippy --all-targets -- -D warnings\` — pass
- \`cargo test --release wallet_keys\` — 55 pass, 0 fail, 1 ignored
- \`cargo test --release paths\` — 5 pass, 0 fail (per-OS resolver, legacy fallback, native-wins-over-legacy)

Full suite intentionally not run (argon2 debug builds are 7+ min per
project test discipline).

## Test plan

- [ ] CI green on linux
- [ ] \`btcli-compat\` workflow still passes against the new location
- [ ] On a macOS dev box: wallet created via \`btt wallet create\` lands at \`\$HOME/Library/Application Support/btt/wallets/<name>/\`
- [ ] On linux with an existing \`\$HOME/.bittensor\`: \`btt wallet list\` emits the legacy-fallback warning once and successfully lists the existing wallets

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,6 +26,40 @@ btt skill
 
 All commands output JSON to stdout. Use `--pretty` for human-readable formatting.
 
+## Config and wallet directory
+
+btt stores wallets and other per-user state under a single OS-dependent
+config directory:
+
+| OS      | Path                                                      |
+| ------- | --------------------------------------------------------- |
+| linux   | `$XDG_CONFIG_HOME/btt` if set, else `$HOME/.config/btt`   |
+| macOS   | `$HOME/Library/Application Support/btt`                   |
+| windows | `%APPDATA%\btt`                                           |
+
+Wallets live at `<config_dir>/wallets/<wallet_name>/` and hold a `coldkey`,
+`coldkeypub.txt`, and `hotkeys/<hotkey_name>` by the same layout that
+btcli uses.
+
+### Legacy path fallback
+
+Earlier versions of btt stored wallets at `$HOME/.bittensor/wallets/` (the
+btcli location). If that directory still exists on disk and the new
+config directory does not, btt continues to read and write the legacy
+location so existing wallets keep working, and prints a one-time warning
+to stderr the first time a command resolves the path:
+
+```
+btt: legacy wallet directory at /home/alice/.bittensor detected.
+     Move it to /home/alice/.config/btt to use the new location:
+         mv /home/alice/.bittensor /home/alice/.config/btt
+     btt will continue to use the legacy location until the move is performed.
+```
+
+btt never moves wallet material on your behalf. Run the `mv` yourself
+when you are ready — if you keep a parallel btcli install, you may prefer
+to leave the legacy location in place and let both tools share it.
+
 ## Non-interactive automation
 
 Commands that prompt for the coldkey password (`wallet create`, `wallet
--- a/scripts/btcli-compat/check.py
+++ b/scripts/btcli-compat/check.py
@@ -202,6 +202,15 @@ class BttDriver:
         self.home = home
         self.verbose = verbose
         self.home.mkdir(parents=True, exist_ok=True)
+        # Pin XDG_CONFIG_HOME so btt's OS-dependent config resolver lands
+        # inside our sandbox on linux. On other platforms we fall through
+        # to HOME-relative paths (macOS: $HOME/Library/..., windows:
+        # %APPDATA%), which btcli-compat does not run on today. The CI
+        # runners are linux-only.
+        self.xdg_config_home = self.home / "xdg"
+        self.xdg_config_home.mkdir(parents=True, exist_ok=True)
+        # Resolved btt config root for this sandbox.
+        self.btt_config_dir = self.xdg_config_home / "btt"
 
     def _run(
         self,
@@ -212,6 +221,7 @@ class BttDriver:
         env = {
             **os.environ,
             "HOME": str(self.home),
+            "XDG_CONFIG_HOME": str(self.xdg_config_home),
             **(env_extra or {}),
         }
         cmd = [str(self.binary), *args]
@@ -287,7 +297,9 @@ class BttDriver:
         return json.loads(proc.stdout.decode("utf-8"))
 
     def coldkey_path(self, wallet_name: str) -> Path:
-        return self.home / ".bittensor" / "wallets" / wallet_name / "coldkey"
+        # Matches `crate::commands::paths::config_dir` on linux with
+        # XDG_CONFIG_HOME set.
+        return self.btt_config_dir / "wallets" / wallet_name / "coldkey"
 
 
 # ── Password file helper (tmpfs-aware, mode 0600) ───────────────────────
@@ -462,7 +474,7 @@ def run_direction_b(
     the inner JSON. btt builds it; we just repack the envelope.
     """
     wallet_name = f"py2btt-{vector.index}"
-    wallet_dir = driver.home / ".bittensor" / "wallets" / wallet_name
+    wallet_dir = driver.btt_config_dir / "wallets" / wallet_name
 
     # Step 1: let btt create the wallet with a known password.
     bootstrap_pw = b"bootstrap-password-not-secret"
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -61,7 +61,11 @@ pub enum ChainAction {
 
 #[derive(Subcommand, Debug)]
 pub enum WalletAction {
-    /// List wallets in ~/.bittensor/wallets/
+    /// List wallets in the btt config directory (linux:
+    /// $XDG_CONFIG_HOME/btt/wallets or $HOME/.config/btt/wallets; macOS:
+    /// $HOME/Library/Application Support/btt/wallets; windows:
+    /// %APPDATA%\btt\wallets). Legacy $HOME/.bittensor/wallets is used
+    /// automatically if present and the new location is not.
     List,
 
     /// Create a new wallet (coldkey + hotkey pair)
--- a/src/commands/mod.rs
+++ b/src/commands/mod.rs
@@ -1,5 +1,6 @@
 pub mod chain;
 pub mod password_file;
+pub mod paths;
 pub mod skill;
 pub mod stake;
 pub mod wallet;
--- a/src/commands/password_file.rs
+++ b/src/commands/password_file.rs
@@ -10,11 +10,11 @@
 //! ----------------
 //! - On unix, we refuse to read a file whose mode is group- or world-readable
 //!   (`mode & 0o077 != 0`). This check is not TOCTOU hardening — the expected
-//!   password file lives at `$HOME/.btt/pwd`, mode 0600, inside a 0700 parent,
-//!   and that posture already closes the TOCTOU window: any file reachable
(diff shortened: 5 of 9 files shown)


moorkh (commented):
[cryptid] Verdict: APPROVE

Hostile review of 52c25e3 (os-dependent config path). All CI gates pass locally; no new dependencies; cross-platform cfg arms are balanced; the legacy-fallback semantics are correct and do not pollute stdout JSON consumers.

## Gates (rustc 1.94.1)

- `cargo build`: pass
- `cargo clippy --all-targets -- -D warnings`: pass (clean, no warnings)
- `cargo test --release paths`: 5/5 pass
- `cargo test --release wallet_keys`: 55/55 pass (1 ignored, pre-existing)
- `git diff main..HEAD -- Cargo.toml Cargo.lock`: empty — stdlib-only confirmed

## Adversarial walk

**cfg arm coverage (paths.rs:43-84).** Four arms: linux, macos, windows, and `not(any(linux, macos, windows))`. All return `Result<PathBuf, BttError>`. The fallback arm mirrors the linux XDG logic verbatim. A typo or missing arm would silently break BSD/illumos; nothing's missing.

**HOME/APPDATA unset error paths.** Each arm propagates `BttError::io("HOME environment variable not set")` (or `APPDATA`) via `map_err`. No panics. Collapsing `VarError::NotUnicode` into the same "not set" message is imprecise but not load-bearing — user-facing text only. Low.

**Empty-string XDG.** `if !xdg.is_empty()` guards `""` explicitly (test `linux_empty_xdg_is_ignored` covers it). Whitespace-only (`" "`) or relative (`"./config"`) falls through to `PathBuf::from(xdg).join("btt")` and then to the caller's `create_dir_all`, which will either succeed as a relative path or fail cleanly — no crash, no undefined behavior. Acceptable.

**Legacy warning → stdout JSON.** `warn_legacy` uses `eprintln!` exclusively (paths.rs:89-104). `stdout` is never touched. Consumers piping `btt wallet list | jq` are unaffected. Good.

**`Once` semantics (paths.rs:87).** `static LEGACY_WARNING: Once = Once::new()` is process-static. Production binary gets a fresh `Once` per invocation — warning fires on the first legacy-resolved call and nowhere else. Correct.

**`ENV_LOCK` typing (paths.rs:154).** `pub(crate) static ENV_LOCK: std::sync::Mutex<()> = std::sync::Mutex::new(())`. `Mutex::new` has been `const` since Rust 1.63; project is rust edition 2021 on 1.94. Fine. `wallet_keys::tests::HOME_LOCK` is a `&'static Mutex<()>` alias onto `ENV_LOCK`, which means both test modules serialize on the same lock as advertised.

**Permissions chain.** `config_dir() → wallets_dir() → ensure_wallets_root() → ensure_secure_dir()` still applies `0o700` via `fs::set_permissions(path, Permissions::from_mode(0o700))` at wallet_keys.rs:914. No relaxation vs. the old `$HOME/.bittensor` behavior. Note: in the legacy-fallback case, `ensure_wallets_root` will chmod the user's `.bittensor` dir to 0o700 just as before — unchanged behavior, but worth mentioning for btcli-side operators.

**Dir-creation on deep macOS paths.** `fs::create_dir_all($HOME/Library/Application Support/btt)` handles the space-containing intermediate. README and help text both show the path with the space correctly (`Library/Application Support`).

**`btcli-compat/check.py` sandboxing.** `BttDriver.__init__` pre-creates `home/xdg` and `home/xdg/btt`. `_run` builds env as `{**os.environ, HOME=..., XDG_CONFIG_HOME=..., **extra}` — a fresh dict per subprocess call, no mutation of parent env. `coldkey_path` and the `run_direction_b` wallet_dir both resolve via `self.btt_config_dir`. No stale `.bittensor` references remain in the script. Linux-only CI runner matches the sandbox's linux-only XDG pin.

**No `.unwrap()` on real paths.** Only in `#[cfg(test)]` blocks; production code uses `map_err` throughout paths.rs and the refactored `wallet_keys.rs` path helpers.

**README & skill.rs accuracy.** Per-OS table matches the code, macOS path has the space, legacy warning example in README quotes the exact `eprintln!` output format (spot-checked against `warn_legacy`).

## Non-blocking nits (follow-up only)

1. **`seat_home_env` sets `XDG_CONFIG_HOME` but wallet_keys test callers only save/restore `HOME`.** After any test that calls `seat_home_env`, `XDG_CONFIG_HOME` is leaked into the process. Because every env-touching test acquires `HOME_LOCK`/`ENV_LOCK` *and* unconditionally re-sets the vars it cares about, the leak is benign in practice. Still, hygiene. A small `XdgGuard` analogous to the `EnvGuard` used in paths.rs tests would close it.

2. **`legacy_fallback_when_only_old_dir_exists` asserts the returned path but not the stderr emission.** If a refactor silently drops the `warn_legacy(...)` call, the test still passes. A cheap improvement: spawn a subprocess or wrap stderr via `gag::BufferRedirect` (dev-dep — don't add) and assert the message contains "legacy wallet directory". At minimum, add a compile-time check that `warn_legacy` is still referenced in `config_dir`.

3. **No test covers the "fired at most once" `Once` guarantee.** `static LEGACY_WARNING: Once = Once::new()` is process-static, so a test that calls `config_dir()` twice in the legacy branch cannot easily re-assert emission — but it could count observed lines on captured stderr. Low priority; the `Once` idiom is a well-known pattern.

4. **Duplicated cfg logic** between `paths::native_config_dir` and the test helper `test_wallets_parent` / `seat_home_env` in wallet_keys. If the resolver's rules ever change, these three sites must all be updated. Consider exposing a `#[cfg(test)] pub(crate) fn expected_resolved_dir(home: &Path) -> PathBuf` in paths.rs so the rules live in one place. Cosmetic; drive-by.

5. **`VarError::NotUnicode` vs. `NotPresent`** collapse into the same "not set" error message. Honest, but less precise than "HOME is not valid UTF-8". Linux allows non-UTF-8 env vars. Incredibly low-probability for `HOME`/`APPDATA`/`XDG_CONFIG_HOME`. Cosmetic.

None of these block the merge. The code is well-factored, genuinely stdlib-only, and the legacy-fallback UX (warn once, never auto-migrate) is the right choice for a tool whose users may also run btcli.

Verdict: APPROVE.

(self-PR, so no --approve flag; this is an approving review comment.)

PR #38, opened 2026-04-14
[cryptid] ci: pin rust toolchain and add --all-features to clippy gate (#33)

moorkh:
[cryptid]

Closes #33.

## Summary

Two LOW findings from the PR #32 barbarian deferred to issue #33:

1. **\`--all-features\` was dropped from the clippy command.** No-op today (\`Cargo.toml\` has no \`[features]\` table) but future-fragile: the moment a feature is added, the gate would silently stop checking the off-by-default code path.
2. **\`dtolnay/rust-toolchain@stable\` floats** in every workflow. A new stable Rust release with a new default clippy lint would simultaneously break every CI workflow — three jobs failing at once with no warning and no clean revert path.

Both fixed here.

## Changes

- **New \`rust-toolchain.toml\`** at the repo root pins channel \`1.94.1\`, includes \`clippy\` and \`rustfmt\` components, uses the \`minimal\` profile. \`cargo\` picks this up on local dev so contributors get the same toolchain as CI without an explicit \`rustup install\` step. The file documents the bump procedure inline.
- **\`.github/workflows/lint.yml\`**: pin to \`dtolnay/rust-toolchain@1.94.1\` and add \`--all-features\` to the clippy command. New comment explains why \`--all-features\` is a no-op today and a guarantee tomorrow.
- **\`.github/workflows/btcli-compat.yml\`**: pin to \`@1.94.1\`.
- **\`.github/workflows/dep-audit.yml\`**: pin to \`@1.94.1\`.

Each pin is annotated with a comment instructing the next bumper to update both the workflow AND \`rust-toolchain.toml\` together. Splitting the bump across PRs would briefly produce a divergence between local and CI toolchains, which is exactly what this issue was filed to prevent.

## Files changed

- \`rust-toolchain.toml\` (new, 17 lines)
- \`.github/workflows/lint.yml\` (+9 / −2)
- \`.github/workflows/btcli-compat.yml\` (+2 / −1)
- \`.github/workflows/dep-audit.yml\` (+2 / −1)

Net: 4 files, +28 / −4. **No source code touched. No new dependencies. No behavior change.**

## Verification (cryptid host)

- \`python3 -c \"import yaml; yaml.safe_load(...)\"\` parses all three workflow files.
- \`rustup\` auto-synced 1.94.1 from the new \`rust-toolchain.toml\` on the first \`cargo\` invocation.
[...]

PR #39, opened 2026-04-14
[cryptid] password_file: bound reads via Read::take and strip UTF-8 BOM (#27, #28)

### LINKED ISSUE #27: [cryptid] password_file: bound read via Read::take(MAX_FILE_BYTES) instead of only checking metadata length
## Context

Raised as a LOW finding during the PR #21 barbarian review (merge commit 381b70a). Filing as a follow-up so the main PR could merge.

## The hole

`read_password_from_file` in `src/commands/password_file.rs` caps the password file at `MAX_FILE_BYTES` (64 KiB) by checking `file.metadata()?.len()` before calling `read_to_end`. That check is advisory: if the file grows between the metadata read and the `read_to_end` call, \`Vec::read_to_end\` just keeps extending. There is no hard ceiling on the read.

On this code path specifically, exploitation is not realistic: the file is 0600 owned by the current uid, so an attacker cannot be writing to it unless they already have the same uid. But the defensive fix is one line and removes a class of footgun:

```rust
// Before:
let mut buf = Vec::new();
file.read_to_end(&mut buf)?;

// After:
let mut buf = Vec::new();
file.take(MAX_FILE_BYTES).read_to_end(&mut buf)?;
// If the file was exactly MAX_FILE_BYTES, it's at the cap — fine.
// If it was more, the read silently truncates — we should then
// check that the remaining file isn't longer and error out.
```

Better shape:

```rust
let mut buf = Vec::with_capacity(md.len().min(MAX_FILE_BYTES) as usize);
let mut reader = file.take(MAX_FILE_BYTES + 1);
reader.read_to_end(&mut buf)?;
if buf.len() as u64 > MAX_FILE_BYTES {
    return Err(BttError::invalid_input(format!(
        "password file at {} exceeds {} bytes (possibly growing under us)",
        path.display(), MAX_FILE_BYTES
    )));
}
```

That reads at most `MAX_FILE_BYTES + 1` bytes and errors if the file is larger than the cap, independent of whatever `metadata()` reported.

## Acceptance

- `read_password_from_file` uses `Read::take` to impose a hard ceiling on the read.
- A test is added that writes a file larger than `MAX_FILE_BYTES` and asserts the caller gets an error rather than a truncated password.
- No new deps.

### LINKED ISSUE #28: [cryptid] password_file: strip UTF-8 BOM and document the footgun
## Context

Raised as an INFO finding during the PR #21 barbarian review (merge commit 381b70a).

## The footgun

PowerShell's `Out-File` (and some other Windows editors) writes UTF-8 files with a leading BOM (\xef\xbb\xbf). If a user creates their password file with PowerShell:

```powershell
"my-password" | Out-File -Encoding utf8 pwd.txt
```

the file contains \xef\xbb\xbf + the password bytes. btt reads this literally — the BOM bytes are prepended to the password that argon2 derives from, so the user's derived key will not match the one they created the wallet with. The user sees a wrong-password error with no obvious way to debug it.

## Proposed fix

1. In `read_password_from_file`, strip a leading UTF-8 BOM if present before stripping trailing newlines.
2. Add a short paragraph to the `--password-file` section of the README noting the BOM hazard, and that PowerShell users should use `Set-Content -Encoding ascii` or explicitly not emit a BOM.
3. Add a test that asserts a password file prefixed with the BOM reads as the bytes after the BOM, not the bytes including it.

No dep on windows detection; just check-and-strip the BOM bytes unconditionally.

## Acceptance

- BOM is stripped on read.
- Test asserts the behavior.
- README documents the case.

moorkh:
## Summary

Two small defensive fixes on `src/commands/password_file.rs`, landed together because the diff is ~30 lines and both touch the same function (`read_password_file_inner`). Both were opened as PR #21 follow-ups.

Closes #27.
Closes #28.

### #27: TOCTOU on the size gate (LOW, not exploitable)

`read_password_file_inner` caps the read at `MAX_FILE_BYTES` (64 KiB) by checking `md.len()` *before* `read_to_end`. That check is advisory: if the file grows between `metadata()` and the read, `Vec::read_to_end` keeps extending and there is no hard ceiling. Not exploitable on this code path — the file is 0600 owned by the caller's own uid, so an attacker capable of growing it already has the caller's privileges — but the defensive fix is one line and removes a class of footgun.

### #28: leading UTF-8 BOM silently prepended to the password (INFO)

PowerShell's `Out-File -Encoding utf8` emits UTF-8 files with a leading BOM (`\xef\xbb\xbf`). btt currently reads these literally, so argon2 sees `BOM || password`, the derived key does not match the one used at wallet creation, and the user gets a wrong-password error with no obvious way to debug it.

## Fixes

- `src/commands/password_file.rs` — switch to `(&mut file).take(MAX_FILE_BYTES + 1).read_to_end(&mut buf)` and error with `BttError::invalid_input` if `buf.len() > MAX_FILE_BYTES`. The advisory `md.len()` gate is preserved as a fast path, but the `Read::take` path is now the authoritative ceiling.
- `src/commands/password_file.rs` — strip a leading `\xef\xbb\xbf` via `buf.drain(..3)` before the existing trailing-newline / first-line extraction logic. Byte-level `starts_with` so no utf-8 validation is run first, and only the leading BOM is special (a BOM later in the buffer is left alone).
- `README.md` — two sentences under `## Non-interactive automation` noting the 64 KiB cap and the BOM behavior, with a pointer to `Set-Content -Encoding ascii` as the preferred PowerShell invocation.
- `src/commands/skill.rs` — matching paragraph under `#### \`--password-file\`` so the three documentation locations (README, skill.rs, `--help`) stay in sync per the no-hidden-flags principle.

## Tests (4 new, all pre-existing still pass)

All tests under `cargo test --release password_file` — 17 passed, 0 failed.

New:
- `oversize_file_errors` — writes `MAX_FILE_BYTES + 100` bytes, asserts the caller gets an error (not a truncated password).
- `at_size_limit_succeeds` — writes exactly `MAX_FILE_BYTES` bytes (`"x\n"` + filler), asserts `"x"` is returned. Exercises the `take(MAX + 1)` path returning exactly `MAX`.
- `bom_is_stripped` — writes `\xef\xbb\xbfpassword\n`, asserts `"password"`.
- `bom_only_at_start` — writes `password\xef\xbb\xbf\n`, asserts the mid-buffer BOM bytes are preserved in the returned password.

Pre-existing tests (13) still pass: `reads_correctly_permissioned_file`, `strips_lf`, `strips_crlf`, `ignores_content_past_first_newline`, `accepts_no_trailing_newline`, `rejects_world_readable`, `rejects_group_readable`, `rejects_missing_file`, `rejects_directory`, `accepts_symlink_to_mode_0600_file`, `expand_tilde_absolute_untouched`, `expand_tilde_relative_untouched`, `expand_tilde_prefix`.

## Verification

- `cargo build` — clean.
- `cargo clippy --all-targets --all-features -- -D warnings` — clean (PR #38 CI gate).
- `cargo test --release password_file` — 17 passed, 0 failed (round 1); 20 passed, 0 failed (round 2).

## Non-goals honored

- No new dependencies.
- No refactor of `read_password_file_inner` beyond the two fixes.
- No changes to any other file under `src/commands/` besides `skill.rs` documentation.
- No changes to the encryption/decryption code (PR #34 territory).

## Round 2 (commit `871e262`)

Barbarian returned REQUEST CHANGES with two LOW findings. Both addressed here; no rebase, no amend.

### LOW-1: BOM-only file previously reached argon2 with an empty password

A file containing only `\xef\xbb\xbf`, only `\xef\xbb\xbf\n`, or zero bytes previously returned `Ok("")` from `read_password_file`, handing argon2 a zero-length secret and deriving a deterministic, attacker-known key. The round-1 BOM-drain path made this newly reachable via BOM-only input; the pre-existing empty-file case was buggy in the same way.

Fix: after the BOM drain and first-line extraction, reject `first_line.is_empty()` with `BttError::invalid_input` naming the offending path. The error message contains the string `"empty"` so tests and users can grep for it.

New tests:
- `bom_only_file_errors` — writes exactly `\xef\xbb\xbf`, asserts error contains `"empty"` or `"BOM"`.
- `bom_then_newline_only_errors` — writes `\xef\xbb\xbf\n`, asserts error contains `"empty"` or `"BOM"`.
- `empty_file_errors` — writes zero bytes, pins the behavior for the pre-existing case.

### LOW-2: `exceeds` branch was masked in `oversize_file_errors`

The `oversize_file_errors` test accepted either `"larger than"` or `"exceeds"` in the error message, masking that only the metadata-time advisory gate was exercised (the file is oversize at stat time, so the `md.len() > MAX_FILE_BYTES` check fires first). The read-time `exceeds` branch is reachable only if the file grows between `metadata()` and `read_to_end()`, which is not achievable from a unit test in safe Rust without wrapping the reader in a custom `Read` impl (scope creep — would require refactoring `read_password_file_inner` to accept a `Read`).

Fix:
- Tightened `oversize_file_errors` to assert specifically `"larger than"` and documented it as exercising the metadata-time check.
- Added a comment block on the read-time `if buf.len() as u64 > MAX_FILE_BYTES` branch explaining it is TOCTOU defense-in-depth, unit-test-unreachable in safe Rust, and that the property it enforces is guaranteed by `Read::take(MAX_FILE_BYTES + 1)` regardless.

### Cosmetic

- Dropped the dead `.min(MAX_FILE_BYTES)` on the buffer-capacity computation. The advisory gate above already rejects `md.len() > MAX_FILE_BYTES`, so `md.len()` is already bounded and the `.min()` was redundant.
- Documented the single-BOM-strip behavior: a file beginning with `\xef\xbb\xbf\xef\xbb\xbf` keeps the second BOM as part of the password. PowerShell never emits a double BOM, and eating arbitrary numbers of leading BOMs would let a crafted file coerce multiple distinct byte sequences into the same derived key.

### Round 2 verification

- `cargo build` — clean.
- `cargo test --release password_file` — **20 passed, 0 failed** (17 from round 1 + 3 new).

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,9 +75,15 @@ btt wallet create --name compat-test --password-file /dev/shm/btt-pw
 shred -u /dev/shm/btt-pw
 ```
 
-On unix, btt refuses to read the file if its mode is other-readable. Do not
-use `--password-file` with mainnet wallets unless your filesystem, process
-listing, and shell history are all under your control.
+On unix, btt refuses to read the file if its mode is other-readable. The
+file must be at most 64 KiB; anything larger is refused outright. A leading
+UTF-8 BOM (`\xef\xbb\xbf`) is stripped on read, so password files created by
+PowerShell's `Out-File -Encoding utf8` (which prepends a BOM) still match
+the password used at wallet creation — but prefer `Set-Content -Encoding
+ascii` or equivalent to avoid the ambiguity in the first place.
+
+Do not use `--password-file` with mainnet wallets unless your filesystem,
+process listing, and shell history are all under your control.
 
 ## Overwrite protection
 
--- a/src/commands/password_file.rs
+++ b/src/commands/password_file.rs
@@ -131,9 +131,36 @@ fn read_password_file_inner(path: &Path) -> Result<Zeroizing<String>, BttError> 
         )));
     }
 
-    let mut buf = Zeroizing::new(Vec::<u8>::with_capacity(md.len() as usize));
-    file.read_to_end(&mut buf)
+    // Enforce a HARD ceiling on the read itself, independent of the advisory
+    // `metadata().len()` check above. If the file grows between the stat and
+    // the read (a TOCTOU that's not exploitable here — the file is 0600,
+    // owned by the caller's uid — but removing the footgun is one line),
+    // `Read::take` will stop at `MAX_FILE_BYTES + 1` and we error out.
+    // Issue #27.
+    let cap = md.len().min(MAX_FILE_BYTES) as usize;
+    let mut buf = Zeroizing::new(Vec::<u8>::with_capacity(cap));
+    (&mut file)
+        .take(MAX_FILE_BYTES + 1)
+        .read_to_end(&mut buf)
         .map_err(|e| BttError::io(format!("failed to read password file: {}", e)))?;
+    if buf.len() as u64 > MAX_FILE_BYTES {
+        return Err(BttError::invalid_input(format!(
+            "password file {} exceeds {} bytes (possibly growing under us); refusing to read",
+            path.display(),
+            MAX_FILE_BYTES
+        )));
+    }
+
+    // Strip a leading UTF-8 BOM if present. PowerShell's `Out-File -Encoding
+    // utf8` emits a BOM; without this, the BOM bytes get prepended to the
+    // password fed to argon2 and the user sees a "wrong password" error with
+    // no obvious cause. Byte-level check so no utf-8 validation runs first.
+    // Only the leading BOM is special; a BOM later in the file is left alone.
+    // Issue #28.
+    const UTF8_BOM: &[u8] = &[0xef, 0xbb, 0xbf];
+    if buf.starts_with(UTF8_BOM) {
+        buf.drain(..UTF8_BOM.len());
+    }
 
     // Take everything up to the first `\n`.
     let first_line_end = buf.iter().position(|&b| b == b'\n').unwrap_or(buf.len());
@@ -290,6 +317,79 @@ mod tests {
 
         fs::remove_file(&link).ok();
         fs::remove_file(&target).ok();
+    }
+
+    // ---- issue #27: hard ceiling via Read::take ----
+
+    #[cfg(unix)]
+    #[test]
+    fn oversize_file_errors() {
+        // File of MAX_FILE_BYTES + 100. The advisory `md.len()` gate catches
+        // this on the first check, but the Read::take path is also exercised:
+        // even if the gate were somehow bypassed (file growing under us), the
+        // take(MAX+1) + len check would still error.
+        let p = tmp_path("oversize");
+        let n: usize = 64 * 1024 + 100;
+        let payload = vec![b'a'; n];
+        write_mode(&p, &payload, 0o600);
+        let err = read_password_file_inner(&p).expect_err("should refuse");
+        assert!(
+            err.message.contains("larger than")
+                || err.message.contains("exceeds"),
+            "msg: {}",
+            err.message
+        );
+        fs::remove_file(&p).ok();
+    }
+
+    #[cfg(unix)]
+    #[test]
+    fn at_size_limit_succeeds() {
+        // File of exactly MAX_FILE_BYTES. The first byte is the password
+        // 'x' followed by '\n', then filler. We should get "x" back and no
+        // size error. Exercises the take(MAX+1) path returning exactly MAX.
+        let p = tmp_path("atlimit");
+        let n: usize = 64 * 1024;
+        let mut payload = Vec::with_capacity(n);
+        payload.extend_from_slice(b"x\n");
+        payload.resize(n, b'y');
+        write_mode(&p, &payload, 0o600);
+        let pw = read_password_file_inner(&p).expect("read at limit");
+        assert_eq!(&*pw, "x");
+        fs::remove_file(&p).ok();
+    }
+
+    // ---- issue #28: strip leading UTF-8 BOM ----
+
+    #[cfg(unix)]
+    #[test]
+    fn bom_is_stripped() {
+        // PowerShell's `Out-File -Encoding utf8` writes \xef\xbb\xbf + bytes.
+        // The BOM must not appear in the password we feed to argon2.
+        let p = tmp_path("bom");
+        let mut payload = Vec::new();
+        payload.extend_from_slice(&[0xef, 0xbb, 0xbf]);
+        payload.extend_from_slice(b"password\n");
+        write_mode(&p, &payload, 0o600);
+        let pw = read_password_file_inner(&p).expect("read");
+        assert_eq!(&*pw, "password");
+        fs::remove_file(&p).ok();
+    }
+
+    #[cfg(unix)]
+    #[test]
+    fn bom_only_at_start() {
+        // A BOM later in the file is a real character and must NOT be
+        // stripped. Only a leading BOM is special.
+        let p = tmp_path("bom-mid");
+        let mut payload = Vec::new();
+        payload.extend_from_slice(b"password");
+        payload.extend_from_slice(&[0xef, 0xbb, 0xbf]);
+        payload.extend_from_slice(b"\n");
+        write_mode(&p, &payload, 0o600);
+        let pw = read_password_file_inner(&p).expect("read");
+        assert_eq!(pw.as_bytes(), b"password\xef\xbb\xbf");
+        fs::remove_file(&p).ok();
     }
 
     #[test]
--- a/src/commands/password_file.rs
+++ b/src/commands/password_file.rs
@@ -131,18 +131,27 @@ fn read_password_file_inner(path: &Path) -> Result<Zeroizing<String>, BttError> 
         )));
     }
 
-    // Enforce a HARD ceiling on the read itself, independent of the advisory
-    // `metadata().len()` check above. If the file grows between the stat and
-    // the read (a TOCTOU that's not exploitable here — the file is 0600,
-    // owned by the caller's uid — but removing the footgun is one line),
-    // `Read::take` will stop at `MAX_FILE_BYTES + 1` and we error out.
-    // Issue #27.
-    let cap = md.len().min(MAX_FILE_BYTES) as usize;
+    // `md.len() > MAX_FILE_BYTES` was rejected above, so `md.len()` is
+    // already bounded. We size the buffer from `md.len()` directly — no
+    // `.min(MAX_FILE_BYTES)` needed. (If that gate is ever removed or
+    // reordered, the `Read::take` below remains the authoritative ceiling.)
+    let cap = md.len() as usize;
     let mut buf = Zeroizing::new(Vec::<u8>::with_capacity(cap));
     (&mut file)
         .take(MAX_FILE_BYTES + 1)
         .read_to_end(&mut buf)
         .map_err(|e| BttError::io(format!("failed to read password file: {}", e)))?;
+    // TOCTOU defense-in-depth: this branch only fires if the file grew
+    // between `metadata()` and `read_to_end()` — the advisory `md.len()`
+    // gate above has already rejected files that were oversize at stat
+    // time. In safe Rust backed by a real filesystem `File`, there is no
+    // way to reach this branch from a unit test without wrapping the
+    // reader in a custom `Read` impl (which would require refactoring
+    // `read_password_file_inner` to accept a `Read` — out of scope).
+    // The property is enforced by `Read::take(MAX_FILE_BYTES + 1)`
+    // regardless of whether this branch is ever unit-test-reachable:
+    // `read_to_end` can never push more than `MAX_FILE_BYTES + 1` bytes
+    // into `buf`, so the comparison is both necessary and sufficient.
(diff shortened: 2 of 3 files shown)


moorkh (commented):
[cryptid] REQUEST CHANGES — two real findings, one LOW behavioural regression and one test-coverage gap. Build/clippy/tests all green; three nits.

## Verdict
**REQUEST CHANGES** — not blocking, but the BOM-strip empty-password regression deserves a tiny guard before merge, and the new `exceeds` branch is dead in the test suite.

## Findings

### LOW-1 — BOM-only / BOM+newline file now reads as an empty password
`src/commands/password_file.rs:161` strips the leading BOM before the first-line split. Consequence:

| file bytes | before PR #39 | after PR #39 |
|---|---|---|
| `\xef\xbb\xbf` (BOM only) | `Ok("\u{feff}")` (utf8 parse of the BOM char) | `Ok("")` |
| `\xef\xbb\xbf\n` | `Ok("\u{feff}")` | `Ok("")` |
| `` (empty file) | `Ok("")` (pre-existing) | `Ok("")` |

Proof (probe test I ran locally against 416869d, then reverted):

```
CRYPTID_BOM_ONLY: Ok("")
CRYPTID_BOM_NL:   Ok("")
CRYPTID_EMPTY:    Ok("")
```

`read_password_file_inner` has no empty-buffer guard. A zero-length password survives all the way to `Zeroizing::new(\"\".to_string())` and gets fed to argon2. On wallet creation that means the coldkey is protected by the empty string; on unlock it means a user who accidentally `> /dev/shm/btt-pw` (redirect truncate) succeeds silently.

The empty-file case is pre-existing and out of scope, but the two BOM cases are new surface area introduced by this PR — previously the BOM bytes at least made the result non-empty. Minimal fix (one line, same function, right after the BOM drain):

```rust
if buf.is_empty() || buf.iter().position(|&b| b == b'\n').unwrap_or(buf.len()) == 0 {
    return Err(BttError::invalid_input(format!(
        \"password file {} is empty (or first line is empty)\",
        path.display()
    )));
}
```

Or just reject empty post-strip and let the existing first-line logic handle the rest. Either way: add a test.

### LOW-2 — new `exceeds` branch is dead in the test suite
The new `buf.len() as u64 > MAX_FILE_BYTES` error at `src/commands/password_file.rs:146` is the authoritative size gate per the PR intent, but it is **never reached** by `oversize_file_errors`. Trace:

1. `oversize_file_errors` writes a `MAX+100` file up-front.
2. `File::open` → `file.metadata()` returns `len = MAX+100`.
3. `md.len() > MAX_FILE_BYTES` at line 126 fires FIRST, returning `\"... is larger than 65536 bytes ...\"`.
4. Control never reaches the `Read::take` path or the new `exceeds` check.

The test asserts the message contains `\"larger than\" || \"exceeds\"`, which masks this — both branches would pass the assertion even though only one is live.

Probe (ran locally, reverted): a `MAX+1`-byte file reports `\"password file ... is larger than 65536 bytes; refusing to read\"` — the advisory message, from line 126, not the new branch. The only path that reaches line 146 is genuine TOCTOU growth (stat returns ≤MAX, then the file grows during read). The test suite does not simulate that.

Fix options (pick one):
- Simulate TOCTOU with a named pipe or by opening-then-writing (awkward, platform-specific).
- Tighten the assertion in `oversize_file_errors` to require the `\"larger than\"` message (documents that this test exercises the advisory gate), and add a comment noting that the `exceeds` branch is defense-in-depth only reachable via file growth under the fd.
- Or add a unit-level test that constructs the buffer directly and asserts on the error type — but that requires refactoring `read_password_file_inner` to split stat from read.

Least-invasive: tighten the existing assertion and add a comment. I would accept that.

## Nits (non-blocking)

**N1** — `src/commands/password_file.rs:140`: `let cap = md.len().min(MAX_FILE_BYTES) as usize;` — the `min` is redundant because line 126 already returned if `md.len() > MAX_FILE_BYTES`. Harmless, but confusing to a reader wondering which branch handles oversize. Drop it or add a `// min is belt-and-suspenders` comment.

**N2** — double-BOM (`\xef\xbb\xbf\xef\xbb\xbfpassword\n`) passes the inner `\u{feff}` through as part of the password silently. PowerShell never emits this, so it's academic, but the code intent (\"strip any number of leading BOMs\" vs \"strip exactly one\") is undocumented. `bom_only_at_start` covers mid-file BOM-preservation but not double-leading-BOM. A one-line comment (\"single strip — PowerShell emits at most one\") would nail down the intent.

**N3** — `src/commands/skill.rs:110` and `README.md:78` use identical wording; they're in sync, good. No action needed, flagging only because the task asked.

## Things I verified and found clean

- `git diff --stat HEAD~1 HEAD` shows only `README.md`, `src/commands/password_file.rs`, `src/commands/skill.rs`. No scope creep.
- `Cargo.toml` / `Cargo.lock` unchanged.
- No `.unwrap()` introduced in production paths.
- `buf.len() as u64` is a widening cast on 32-bit (usize=u32 → u64), safe.
- `at_size_limit_succeeds` really does create a 65536-byte file (`\"x\\n\" + 65534 * 'y'`). Boundary is honest.
- UTF-16 LE/BE BOMs (`\\xff\\xfe`, `\\xfe\\xff`) are not stripped — they trip the `from_utf8` guard and produce `\"does not contain valid UTF-8 on the first line\"`, which is an acceptable failure mode. No action needed.
- `Read::take` boundary is correct: file=MAX → buf.len()=MAX → passes; file=MAX+1 → take yields MAX+1 → `> MAX` fires (in the TOCTOU case where the advisory gate missed it).
- `metadata()` is taken from the open fd (not re-stat'd from the path), so the is_file/mode checks describe the bytes being read. Good hygiene, predates this PR.

## CI gates (local)
```
cargo build                                        -> clean
cargo clippy --all-targets --all-features -D warn  -> clean
cargo test --release password_file                 -> 17 passed, 0 failed (4 new + 13 pre-existing)
```

## Follow-ups

- **Fix LOW-1** in this PR: one-line empty-buffer guard after BOM drain, plus a `rejects_bom_only_file` test.
- **Tighten LOW-2 test assertion** in this PR: require `\"larger than\"` and document that the `exceeds` branch is TOCTOU-defense only.
- **Separate issue** (not for this PR): `read_password_file_inner` silently accepts a 0-byte file as an empty password. That's pre-existing behaviour worth a `rejects_empty_file` test and guard — opening a new issue is cleaner than piggybacking here.

Not merging.

moorkh (commented):
[cryptid] round 2 review of 871e262

**APPROVE** — both round-1 findings are genuinely fixed. Loop terminates.

## Verification performed

Ran in the worktree on `cryptid/password-file-take-and-bom` at 871e262:

- `cargo build` — clean
- `cargo clippy --all-targets --all-features -- -D warnings` — clean
- `cargo test --release password_file` — 20 passed, 0 failed

Diff scope (`git diff --stat 416869d HEAD`) touches only `src/commands/password_file.rs`. `Cargo.toml` / `Cargo.lock` unchanged across both round commits. No new `.unwrap()` / `.expect()` / `panic!` in production paths.

## LOW-1 — empty / BOM-only guard

Traced the call path: `read_password_file` → `read_password_file_inner` → open → metadata checks → read → BOM-strip → first-line extraction → `if first_line.is_empty()` → `from_utf8` → return. The empty-check is the last guard before return. Caller in `main.rs::resolve_coldkey_password` uses `?` and the top-level `run` bubbles to `process::exit(1)`. No swallow path.

End-to-end integration probes against `./target/release/btt wallet new-coldkey --password-file <f>`:

| input bytes                  | result                                                                                              |
|------------------------------|-----------------------------------------------------------------------------------------------------|
| `\xef\xbb\xbf`               | `INVALID_INPUT`: `password file … is empty (or contains only a BOM and/or a trailing newline)`      |
| `\xef\xbb\xbf\n`             | same                                                                                                |
| zero-byte                    | same                                                                                                |
| `\xef\xbb\xbfx` (BOM + `x`)  | success — empty-check does NOT false-fire                                                           |
| bare `x` (one byte, no NL)   | success                                                                                             |

Error message contains `empty`, contains `BOM` (covers both unit-test assertion shapes), and names the absolute path of the offending file. Sufficient to debug.

## LOW-2 — `exceeds` branch documentation and test tightening

- `oversize_file_errors` now asserts specifically `err.message.contains(\"larger than\")` with no `|| contains(\"exceeds\")` fallback. The gap round-1 raised (advisory gate masks the take-path gate) is now visible to the test harness.
- The comment on the `exceeds` branch correctly describes it as TOCTOU defense-in-depth, correctly cites `Read::take(MAX_FILE_BYTES + 1)` as the authoritative ceiling, and correctly explains why it is unit-test-unreachable in safe Rust without a custom `Read` impl.
- I considered whether a fifo / named pipe could exercise the `exceeds` branch (metadata reports 0, reader produces arbitrary bytes). It cannot: the earlier `md.is_file()` check rejects fifos before the read happens. The builder's unreachability claim holds.

## Cosmetic nits

- `cap = md.len() as usize` — dead `.min(MAX_FILE_BYTES)` dropped, comment explains why it is safe.
- BOM-strip block now documents the single-BOM-only behavior with the correct rationale (double-BOM would coerce multiple distinct byte sequences into the same derived key). Accurate.

## Verdict

APPROVE. The round-1 findings are sealed. The patch is tight, the tests pin the behavior the patch claims, and the integration probes confirm the guard fires end-to-end before any key derivation. Merge.

PR #40, opened 2026-04-14
[cryptid] wallet create: stage-and-rename for atomic wallet creation (#29)

### LINKED ISSUE #29: [cryptid] wallet create: transactional half-write gap (pre-existing)
## Context

Raised as a follow-up from the PR #26 barbarian review (merge commit 0f6769f). Pre-existing issue, not a regression from PR #26.

## The gap

`btt wallet create` writes the coldkey file first, then writes the hotkey file. If the hotkey write fails (disk full, permission flip, signal), the wallet directory ends up in a half-written state: coldkey present, hotkey missing or truncated.

The `guard_create_overwrite` helper (PR #26) prevents a subsequent `wallet create --name <same>` from clobbering the partial wallet, but a user in this state has no clean way forward without either:
- Manually deleting the partial wallet directory, or
- Running `wallet create --force` (which they may not realize is needed).

The PR #26 barbarian traced the code path: `src/commands/wallet_keys.rs::create` does

1. generate coldkey
2. write coldkey
3. generate hotkey
4. write hotkey

Steps 2 and 4 are independent file writes. If 4 fails, the wallet is in an inconsistent state with the coldkey committed.

## Proposed fix

One of:
- **Stage into a temp dir and move into place.** Write both files into a `.wallet-create-tmp.<pid>` directory next to the target, then atomically rename the directory. Requires that both writes succeed before anything becomes user-visible.
- **Roll back on hotkey failure.** On step-4 error, delete the coldkey file written in step 2 and re-raise the original error.

Option A is cleaner because it avoids touching the user's wallet directory at all on failure. Option B is simpler but requires care around the `coldkeypub.txt` file too.

## Acceptance

- Kill the mid-create hotkey write via a test fixture and assert the wallet dir is either fully populated or empty, never partial.
- No change to the success path.
- No new deps.

moorkh:
Closes #29.

## Summary

`btt wallet create` previously wrote the three wallet artifacts (coldkey, coldkeypub.txt, hotkey) directly into `<wallets>/<name>/`. A failure between writes left the user with a partial wallet that `guard_create_overwrite` (PR #26) then refused to replace without `--force`. This PR fixes the transactional half-write gap by staging all three files into a sibling temp directory and atomically renaming it into place.

## Before / after

**Before (src/commands/wallet_keys.rs::create):**
```
1. ensure_secure_dir(<wallets>/<name>)
2. write_secure_file(<wallets>/<name>/coldkey)            ← point of no return
3. fs::write(<wallets>/<name>/coldkeypub.txt)
4. ensure_secure_dir(<wallets>/<name>/hotkeys)
5. write_secure_file(<wallets>/<name>/hotkeys/<hotkey>)   ← if this fails → partial wallet
```
If step 5 failed, the wallet dir was half-written (coldkey + coldkeypub present, hotkey missing), and a re-run without `--force` was refused.

**After:**
```
1. guard_create_overwrite(...)                            ← unchanged
2. ensure_secure_dir(<wallets>/.tmp.<name>.<pid>.<nanos>)
3. write_secure_file(<staging>/coldkey)
4. fs::write(<staging>/coldkeypub.txt)
5. ensure_secure_dir(<staging>/hotkeys)
6. write_secure_file(<staging>/hotkeys/<hotkey>)
7. fs::rename(<staging>, <wallets>/<name>)                ← atomic on a single fs
```
Any error in 2–6 triggers `fs::remove_dir_all(<staging>)` and the user's wallet directory is never touched. On Linux a directory `rename(2)` is atomic on the same filesystem, so after step 7 the target is either fully populated or absent — never half-written.

## Temp directory naming scheme

`<wallets>/.tmp.<wallet_name>.<pid>.<nanos>/`

- `.tmp.` prefix: reserved. `wallet list` skips any entry starting with `.tmp.` so stale staging dirs from a crashed run do not surface as real wallets.
- `<pid>`: distinguishes concurrent `create` invocations from different processes.
- `<nanos>`: `SystemTime::now().as_nanos()` — distinguishes repeated calls inside a single process and reduces the chance a stale temp dir from a prior crashed run collides with a fresh one after a PID wrap.

Stale `.tmp.*` directories are **not** auto-cleaned — they are deliberately left as forensics for the user.

## --force semantics

Preserved exactly. When the target dir already exists (only reachable under `--force`, since `guard_create_overwrite` otherwise refuses), the individual staged artifacts are moved into the existing directory one by one via `promote_staged_into_existing`. This is weaker than a directory-level rename atomicity-wise but deliberately preserves the pre-#29 behavior — specifically, unrelated hotkeys under `<wallet>/hotkeys/` that the guard does not name continue to survive the force-overwrite. A directory-level rename would silently wipe them.

## Tests added (4)

All live in `src/commands/wallet_keys.rs::tests`, hold `HOME_LOCK`, and use the existing `seat_home`/`restore_home` helpers:

1. `wallet_create_atomic_on_success` — after a success, the target dir contains exactly `coldkey`, `coldkeypub.txt`, and `hotkeys/<name>`.
2. `wallet_create_no_temp_dir_after_success` — after a success, no `.tmp.*` directory remains under `<wallets>/`.
3. `wallet_create_failure_leaves_no_partial` — injects a synthetic failure via the `BTT_FAIL_AFTER_HOTKEY_WRITE` env var (test-only hook gated on `#[cfg(test)]`), asserts the error carries the synthetic tag, the target dir does not exist, and no `.tmp.*` is left behind.
4. `wallet_list_skips_temp_dirs` — plants a `.tmp.real.99999.123456789/` directory with a dummy `coldkeypub.txt`, runs `wallet::list`, and asserts the listing contains the real wallet but no entry starting with `.tmp.`.

## Failure injection

A real failure injection (e.g. mock writer trait) would have been a larger refactor of `create`'s signature. I instead used a `#[cfg(test)]`-gated env-var hook (`BTT_FAIL_AFTER_HOTKEY_WRITE`) inside `create_into_staging` that, when set, scrubs the in-flight key material and returns an `io` error after the hotkey has been staged. The hook is compiled out of the release binary entirely — `cfg(test)` gates both the `env::var_os` read and the error path, so there is no `#[clap(hide = true)]`-style concealment. The test asserts on the synthetic error tag to make sure no real I/O failure can masquerade as a passing test.

## Test plan

- [x] `cargo build` clean
- [x] `cargo clippy --all-targets --all-features -- -D warnings` clean
- [x] `cargo test --release wallet_keys` — 59 passed, 1 ignored
- [x] `cargo test --release wallet` — 59 passed, 1 ignored (same module, filter is a superset)
- [x] All four new tests assert the atomicity invariants
- [x] No new dependencies
- [x] `CreateResult` API unchanged
- [x] `guard_create_overwrite` unchanged, runs before any staging

---

## Round 2 (commit `6147c4e`)

Round-1 reviewer flagged one HIGH and several smaller issues. This commit lands the fixes.

### HIGH: promote path half-write regression

Rewrote `promote_staged_into_existing` from a three-step sequential file-rename into a backup-and-publish algorithm:

1. `rename(target, .bak.<name>.<pid>.<nanos>.<ctr>)` — move the old wallet aside into a sibling backup dir.
2. `rename(staging, target)` — atomic publish. On failure, `rename(backup, target)` rolls back and the original wallet is restored byte-for-byte.
3. Merge unrelated hotkeys from `backup/hotkeys/` into `target/hotkeys/`, skipping the freshly-published hotkey (the new one wins). This preserves the pre-#29 `--force` semantic that unrelated hotkeys survive a force-overwrite.
4. Best-effort `remove_dir_all(backup)`.

Crash window: between steps 1 and 2 the target name is briefly absent on disk, but the full contents live in `.bak.*`. A crash there leaves the backup reachable for forensics. Between 2 and 3 the new wallet is already live.

### MEDIUM: `wallet_dir.exists()` branch gated on force

`if wallet_dir.exists()` is now `if wallet_dir.exists() && force`. Without the `&& force` gate, a crash-leftover wallet dir containing only unrelated hotkeys (no coldkey, no named hotkey) passed `guard_create_overwrite` and silently fell into the force path — the `debug_assert!(force, ...)` was compiled out in release. The no-force branch now falls through to the plain `fs::rename` which refuses with `ENOTEMPTY` on a non-empty target.

### LOW: `.tmp.`/`.bak.` reserved prefix

`create` now rejects wallet names starting with `.tmp.` or `.bak.` with an `invalid_input` error. `wallet::list` also filters the `.bak.*` prefix (in addition to the existing `.tmp.*`).

### Nits

- `temp_staging_name` and the new `backup_dir_for` share a process-local `AtomicU64` counter so filenames stay unique under a frozen or backward-running clock.
- `promote_staged_into_existing` no longer calls `ensure_secure_dir` on the pre-existing target dir — that would have relaxed a user-tightened `chmod` (e.g. `0500`) back to `0700`.

### New tests (3)

1. `wallet_create_force_atomic_on_failure` — fires a new `BTT_FAIL_DURING_PROMOTE` hook between the atomic publish and the unrelated-hotkey merge. Asserts the new wallet is fully live on disk (coldkey, coldkeypub, new hotkey), the pre-overwrite ss58 is gone from coldkeypub, and the `.bak.*` dir still exists so unrelated hotkeys are recoverable.
2. `wallet_create_force_partial_promote_failure_keeps_old` — fires a new `BTT_FAIL_BEFORE_PUBLISH` hook that forces the publish rename to fail after the backup rename succeeds. Asserts the rollback path restores the original wallet byte-for-byte (coldkey, hotkey, coldkeypub) with the original ss58, and that no `.tmp.*` or `.bak.*` dirs remain.
3. `wallet_create_rejects_reserved_prefix` — `.tmp.foo` and `.bak.foo` names return `invalid_input` naming the reserved prefix.

Both test hooks are `#[cfg(test)]`-gated and compile out of the release binary.

### Deferred to follow-up issues

- **Concurrent `--force` races**: per-wallet `flock(2)` is a bigger change than a round-2 fix. Filed as a separate issue (see below).
- **Stale `.tmp.*`/`.bak.*` reaping**: a `wallet cleanup` subcommand is out of scope for this PR. Filed as a separate issue.

### Round-2 test plan

- [x] `cargo build` clean
- [x] `cargo clippy --all-targets --all-features -- -D warnings` clean
- [x] `cargo test --release wallet_keys` — 62 passed, 1 ignored (3 new tests)
- [x] `cargo test --release wallet` — 62 passed, 1 ignored (same module, filter superset)
- [x] All three new tests exercise the new backup-and-rename algorithm

### DIFF
--- a/src/commands/wallet.rs
+++ b/src/commands/wallet.rs
@@ -55,6 +55,16 @@ pub fn list() -> Result<WalletList, BttError> {
             .to_string_lossy()
             .to_string();
 
+        // Skip staging directories left behind by `wallet create`
+        // (issue #29). The atomic-create path writes a sibling temp dir
+        // named `.tmp.<wallet>.<pid>.<nanos>` inside `<wallets>/` and
+        // renames it into place on success. A stale `.tmp.*` from a
+        // crashed run is left on disk deliberately (forensics), but must
+        // not appear in `wallet list` as if it were a real wallet.
+        if wallet_name.starts_with(".tmp.") {
+            continue;
+        }
+
         // Read coldkeypub.txt
         let coldkey_path = path.join("coldkeypub.txt");
         let coldkey = if coldkey_path.exists() {
--- a/src/commands/wallet.rs
+++ b/src/commands/wallet.rs
@@ -55,13 +55,17 @@ pub fn list() -> Result<WalletList, BttError> {
             .to_string_lossy()
             .to_string();
 
-        // Skip staging directories left behind by `wallet create`
-        // (issue #29). The atomic-create path writes a sibling temp dir
-        // named `.tmp.<wallet>.<pid>.<nanos>` inside `<wallets>/` and
-        // renames it into place on success. A stale `.tmp.*` from a
-        // crashed run is left on disk deliberately (forensics), but must
-        // not appear in `wallet list` as if it were a real wallet.
-        if wallet_name.starts_with(".tmp.") {
+        // Skip staging and backup directories left behind by
+        // `wallet create` (issue #29). The atomic-create path writes a
+        // sibling `.tmp.<wallet>.<pid>.<nanos>.<ctr>` staging dir, and
+        // the `--force` promote path writes a sibling
+        // `.bak.<wallet>.<pid>.<nanos>.<ctr>` backup dir during the
+        // swap. Both prefixes are reserved (see `wallet_keys::create`,
+        // which refuses to create a wallet with either prefix). Stale
+        // `.tmp.*` / `.bak.*` dirs from a crashed run are left on disk
+        // deliberately (forensics) but must not appear in `wallet list`
+        // as if they were real wallets.
+        if wallet_name.starts_with(".tmp.") || wallet_name.starts_with(".bak.") {
             continue;
         }
 
--- a/src/commands/wallet_keys.rs
+++ b/src/commands/wallet_keys.rs
@@ -358,6 +358,17 @@ fn guard_create_overwrite(
 /// and caused a second invocation of `wallet create --name <same>` to
 /// silently destroy an irrecoverable wallet. See also PR #18 (issue #7)
 /// for the same fix on `new-coldkey` / `new-hotkey` / `regen-*`.
+///
+/// Issue #29 (atomic creation): all three wallet artifacts (coldkey,
+/// coldkeypub.txt, hotkey) are staged into a sibling temp directory
+/// `<wallets>/.tmp.<name>.<pid>.<nanos>/` and then moved into place with
+/// a single `fs::rename`. The rename is atomic on a single filesystem
+/// (both paths live under `<wallets>/`), so the target `<wallets>/<name>`
+/// directory is either absent or fully populated — never half-written.
+/// On any error during staging, the temp dir is removed and the user's
+/// wallet directory is never touched. The `.tmp.` prefix is excluded from
+/// `wallet list`, so a stale temp dir from a crashed run remains visible
+/// on disk (for forensics) without leaking into the user-facing listing.
 pub fn create(
     wallet_name: &str,
     hotkey_name: &str,
@@ -378,6 +389,156 @@ pub fn create(
     let hotkey_path = wallet_dir.join("hotkeys").join(hotkey_name);
     guard_create_overwrite(wallet_name, &coldkey_path, &hotkey_path, force)?;
 
+    // Stage into a sibling temp dir so the final rename is atomic.
+    // The temp dir lives inside `<wallets>/` (same filesystem as the
+    // target) and uses the `.tmp.` prefix reserved for staging. The
+    // `<pid>.<nanos>` suffix keeps concurrent `create` invocations from
+    // colliding in the same process and reduces the chance that a stale
+    // temp dir from an earlier crashed run collides with a fresh one
+    // after a PID wrap. Stale temp dirs are deliberately left on disk —
+    // they're intended forensics, not to be auto-cleaned on startup.
+    ensure_wallets_root()?;
+    let wallets_root = paths::wallets_dir()?;
+    let staging_dir = wallets_root.join(temp_staging_name(wallet_name));
+
+    // Drive the staged writes through a helper so the `?` propagation is
+    // funneled through a single cleanup point. On any error, remove the
+    // staging dir and return the original error.
+    let cr = match create_into_staging(
+        wallet_name,
+        hotkey_name,
+        n_words,
+        password,
+        &staging_dir,
+    ) {
+        Ok(v) => v,
+        Err(e) => {
+            let _ = fs::remove_dir_all(&staging_dir);
+            return Err(e);
+        }
+    };
+
+    // Publish. Two paths:
+    //
+    // 1. Target does not exist (the common, non-force case): `fs::rename`
+    //    the staging dir into place. Single filesystem, so the rename is
+    //    atomic at the kernel level — the target is either absent or
+    //    fully populated.
+    //
+    // 2. Target already exists (only reachable with --force, because the
+    //    guard otherwise refuses): move the individual staged artifacts
+    //    into the existing target dir. This is NOT atomic across all
+    //    three files (coldkey, coldkeypub, hotkey), but it deliberately
+    //    preserves the pre-#29 --force semantics — specifically, other
+    //    hotkeys under `<wallet>/hotkeys/` that were not overwritten by
+    //    the guard-named hotkey continue to survive. A directory-level
+    //    `rename` would wipe them, which is a silent behavior change and
+    //    a potential second #19.
+    if wallet_dir.exists() {
+        debug_assert!(
+            force,
+            "target exists without --force: guard_create_overwrite should have refused"
+        );
+        if let Err(e) = promote_staged_into_existing(&staging_dir, &wallet_dir, hotkey_name) {
+            let _ = fs::remove_dir_all(&staging_dir);
+            return Err(e);
+        }
+        // The staged hotkeys/ subdir is left empty once its contents move;
+        // remove the now-empty staging dir so no forensics remain.
+        let _ = fs::remove_dir_all(&staging_dir);
+    } else {
+        if let Err(e) = fs::rename(&staging_dir, &wallet_dir) {
+            let _ = fs::remove_dir_all(&staging_dir);
+            return Err(BttError::io(format!(
+                "failed to atomically publish wallet {} ({} -> {}): {}",
+                wallet_name,
+                staging_dir.display(),
+                wallet_dir.display(),
+                e
+            )));
+        }
+    }
+
+    Ok(cr)
+}
+
+/// --force path: move the three staged artifacts (coldkey, coldkeypub,
+/// hotkey) from `staging_dir` into the pre-existing `target_dir`,
+/// overwriting the prior copies. Leaves any unrelated hotkeys under
+/// `<target>/hotkeys/` untouched, matching the pre-#29 semantics.
+fn promote_staged_into_existing(
+    staging_dir: &Path,
+    target_dir: &Path,
+    hotkey_name: &str,
+) -> Result<(), BttError> {
+    // Ensure the per-wallet and hotkeys subdir exist with the 0700 mode.
+    ensure_secure_dir(target_dir)?;
+    ensure_secure_dir(&target_dir.join("hotkeys"))?;
+
+    let moves: [(PathBuf, PathBuf); 3] = [
+        (staging_dir.join("coldkey"), target_dir.join("coldkey")),
+        (
+            staging_dir.join("coldkeypub.txt"),
+            target_dir.join("coldkeypub.txt"),
+        ),
+        (
+            staging_dir.join("hotkeys").join(hotkey_name),
+            target_dir.join("hotkeys").join(hotkey_name),
+        ),
+    ];
+    for (src, dst) in moves {
+        // Remove any pre-existing destination file so `rename` does not
+        // surprise on EEXIST on platforms where that matters. Linux
+        // rename(2) clobbers regular files, but we keep the explicit
+        // unlink for portability and symmetry with `write_secure_file`.
+        if dst.exists() {
+            fs::remove_file(&dst).map_err(|e| {
+                BttError::io(format!(
+                    "failed to remove existing {} prior to force overwrite: {}",
+                    dst.display(),
+                    e
+                ))
+            })?;
+        }
+        fs::rename(&src, &dst).map_err(|e| {
+            BttError::io(format!(
+                "failed to move staged {} into place at {}: {}",
+                src.display(),
+                dst.display(),
+                e
+            ))
+        })?;
+    }
+    Ok(())
+}
+
+/// Compute the staging directory name for an atomic `create`.
+///
+/// Format: `.tmp.<wallet_name>.<pid>.<nanos>`. The `.tmp.` prefix is
+/// reserved (wallet list ignores it), the pid distinguishes concurrent
+/// processes, and the nanosecond timestamp distinguishes repeated calls
+/// from the same process (and reduces the chance that a stale temp dir
+/// from a prior crashed run collides with a fresh one after a pid wrap).
+fn temp_staging_name(wallet_name: &str) -> String {
+    let pid = std::process::id();
+    let nanos = std::time::SystemTime::now()
+        .duration_since(std::time::UNIX_EPOCH)
+        .map(|d| d.as_nanos())
+        .unwrap_or(0);
+    format!(".tmp.{wallet_name}.{pid}.{nanos}")
+}
+
+/// Stage a complete wallet (coldkey + coldkeypub + hotkey) into
+/// `staging_dir`. Caller owns the lifecycle of `staging_dir`: on any
+/// error returned from this helper, the caller must `remove_dir_all`
+/// the staging dir before surfacing the error.
+fn create_into_staging(
+    wallet_name: &str,
+    hotkey_name: &str,
+    n_words: u32,
+    password: &str,
+    staging_dir: &Path,
+) -> Result<CreateResult, BttError> {
     let (cold_pair, mut cold_phrase, mut cold_seed) = generate_keypair(n_words)?;
     let cold_ss58 = cold_pair.public().to_ss58check();
     let cold_seed_hex = format!("0x{}", hex::encode(cold_seed));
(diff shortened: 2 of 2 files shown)


moorkh (commented):
[cryptid] REQUEST CHANGES (self-PR, posting as comment; do not merge)

Barbarian review of PR #40 (commit `ec96a8a`). The non-force path is clean — stage + `fs::rename` is atomic on a single filesystem, cleanup funnels through one point, error propagation is correct, tests cover the success and the injected-failure shapes. But the `--force` path re-introduces issue #29 in a slightly different dress.

## HIGH — `promote_staged_into_existing` is a half-write on --force

`src/commands/wallet_keys.rs:469-513`. On the `--force` branch the three staged artifacts are moved one at a time in sequence `[coldkey, coldkeypub.txt, hotkeys/<name>]`. If the second or third `rename` fails, the first has already been committed and cannot be rolled back. The user sees the error, assumes nothing happened, and their old coldkey is silently gone — replaced by a freshly-generated one whose mnemonic was never surfaced to them.

This is exactly the half-write #29 was supposed to close, moved from the non-force path to the force path.

### PoC (verified locally)

I reverted after running, but here is the test body I added to `src/commands/wallet_keys.rs::tests`:

```rust
#[test]
fn poc_force_midpromote_failure_corrupts_wallet() {
    let _guard = HOME_LOCK.lock().expect("home lock");
    let (tmp, original) = seat_home("poc-force-corrupt");

    let first = create("victim", "default", 12, "pw", false).expect("first create");
    let wdir = test_wallets_parent(&tmp).join("victim");

    let original_coldkey_bytes =
        std::fs::read(wdir.join("coldkey")).expect("read original coldkey");

    // Sabotage step 2: replace coldkeypub.txt with a non-empty directory
    // so the remove_file in promote_staged_into_existing fails with EISDIR.
    std::fs::remove_file(wdir.join("coldkeypub.txt")).expect("rm");
    std::fs::create_dir(wdir.join("coldkeypub.txt")).expect("plant dir");
    std::fs::write(wdir.join("coldkeypub.txt").join("sentinel"), b"x").expect("sentinel");

    let second = create("victim", "default", 12, "pw", true);

    let post_coldkey_bytes =
        std::fs::read(wdir.join("coldkey")).expect("read post-failure coldkey");

    restore_home(tmp, original);
    let _err = unwrap_err(second, "force mid-promote should fail");
    assert_ne!(
        original_coldkey_bytes, post_coldkey_bytes,
        "first_ss58={}", first.coldkey_ss58,
    );
}
```

Run result against `ec96a8a`:

```
test commands::wallet_keys::tests::poc_force_midpromote_failure_corrupts_wallet ... ok
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 115 filtered out
```

The `assert_ne!` fires — i.e. the coldkey bytes HAVE been replaced even though the `create` call returned `Err`. The old wallet is unrecoverable.

Any kernel-level failure mode that can fail a `rename(2)` mid-sequence reproduces this in the wild: ENOSPC on the filesystem, EDQUOT, EROFS after a remount, EIO, an immutable chattr on `coldkeypub.txt`, SELinux/AppArmor denial on the second file, etc. The PoC uses EISDIR because it is deterministic, but the bug does not need that specific trigger.

### Proposed fix

Treat the force path with the same atomicity discipline as the non-force path. Two options, both viable:

1. **Rename the old wallet out of the way first, then rename staging into place, then merge back unrelated hotkeys.**
   ```
   let backup = wallets_root.join(format!(".bak.{name}.{pid}.{nanos}"));
   fs::rename(&wallet_dir, &backup)?;       // atomic: old wallet preserved
   if let Err(e) = fs::rename(&staging_dir, &wallet_dir) {
       let _ = fs::rename(&backup, &wallet_dir); // try to restore
       return Err(e);
   }
   // Copy any hotkeys from backup/hotkeys/ whose name != hotkey_name
   // into wallet_dir/hotkeys/, then remove_dir_all(&backup).
   ```
   The target is always either the old wallet or the new wallet; never partial. Unrelated-hotkey preservation happens after the atomic swap. Filter `.bak.*` out of `wallet list` the same way `.tmp.*` is filtered.

2. **`renameat2` with `RENAME_EXCHANGE` on linux ≥ 3.15** (libc::renameat2 behind a `#[cfg(target_os = "linux")]`). Atomically swap `staging_dir` and `wallet_dir`; the old wallet contents end up in `staging_dir`; iterate over `staging_dir/hotkeys/` and move unrelated hotkeys back into `wallet_dir/hotkeys/`; `remove_dir_all(&staging_dir)`. Non-linux falls back to option 1.

Either way, the three-file sequence in `promote_staged_into_existing` has to go — it is structurally non-atomic.

The comment at `wallet_keys.rs:429-436` anticipates exactly this concern ("this is NOT atomic across all three files ... it deliberately preserves the pre-#29 --force semantics"), but the conclusion is wrong: the pre-#29 force semantics were themselves subject to issue #29. Preserving unrelated hotkeys does not require sacrificing atomicity — options 1 and 2 preserve both.

## MEDIUM — `wallet_dir.exists()` is not a reliable force indicator

`wallet_keys.rs:437-444`. The branch condition is `if wallet_dir.exists()`, gated by `debug_assert!(force, ...)`. Consider a wallet directory that survives from a crash with only `hotkeys/other/` inside it — no `coldkey`, no `hotkeys/<named>`. `guard_create_overwrite` sees neither named file, returns `Ok` without requiring `force`. `wallet_dir.exists()` is true. In release, the `debug_assert!` is compiled out, and `promote_staged_into_existing` runs with `force=false`. The behavior happens to work here (file moves fall through the `dst.exists()` guards), but it is not what the code claims, and it means the "only reachable with --force" comment at `wallet_keys.rs:428` is false. Tighten the branch to `if wallet_dir.exists() && force` (take the atomic-rename path otherwise — there is nothing to preserve), or make the assertion a real release-mode bail-out.

## MEDIUM — Concurrent `--force` on the same wallet

Two processes simultaneously running `wallet create --force --name foo` both pass the guard, both stage into unique dirs, both call `promote_staged_into_existing`. The three `fs::rename` calls interleave unpredictably. Result: a wallet whose `coldkey` came from process A, `coldkeypub.txt` from process B, `hotkeys/default` from either one. The coldkey and coldkeypub no longer correspond. No file lock in the create path. Option 1 above (rename-into-backup) does not fix this on its own, but a `flock(2)` on `<wallets>/.lock.<name>` closes it. Worth at least a follow-up issue.

## LOW — `wallet list` starts_with filter shadows user wallet names

`wallet.rs:58-65`. `starts_with(".tmp.")` hides any user-named wallet that the user deliberately (or accidentally) created as `.tmp.something`. The dot-prefix convention makes this defensible, but the wallet name is not currently validated against the reserved prefix. Consider either (a) rejecting wallet names that start with `.tmp.` in `wallet create` (and `new-coldkey`), or (b) documenting the reserved prefix in the `wallet create --help` text. A comment in `wallet.rs` pointing at the staging-dir contract would also help the next person who touches this filter.

## LOW — Stale `.tmp.*` disk usage accumulates across crashes

PR body says this is intended forensics. Fine, but there is no `wallet cleanup` command and no upper bound. A loop of crashed creates eventually grows `<wallets>/` unboundedly. Follow-up issue: `btt wallet cleanup` that prompts-to-delete any `.tmp.*` dir older than N hours, and (with the fix above) any `.bak.*` dir.

## Nits

- `temp_staging_name` uses `SystemTime::now().duration_since(UNIX_EPOCH).map(...).unwrap_or(0)`. If the clock is earlier than `UNIX_EPOCH` (e.g. a misconfigured VM) two concurrent calls both collapse to `.tmp.<name>.<pid>.0` and collide. Prefer a `u128` process-local counter fallback or a RNG tiebreaker.
- `promote_staged_into_existing` calls `ensure_secure_dir(target_dir)` on an existing wallet dir, which re-chmods it to 0700. If the user had intentionally tightened it further (e.g. 0500), that tightening is relaxed. Minor.
- The `wallet_create_failure_leaves_no_partial` test only covers the failure-injection point that is *after* the hotkey write. Failures earlier in `create_into_staging` (e.g. `ensure_secure_dir(staging/hotkeys)` failing) are inferred from the same cleanup code path but never directly exercised. Consider a second injection point or a second test that tampers with the staging dir's permissions mid-create.

## Scope / hygiene

- `git diff --stat HEAD~1 HEAD` — only `src/commands/wallet.rs` and `src/commands/wallet_keys.rs`. Clean. No Cargo changes.
- No new `.unwrap()` on real failure paths; the `unwrap_or(0)` in `temp_staging_name` is the only one and it is non-secret.
- `#[cfg(test)]` correctly gates the `BTT_FAIL_AFTER_HOTKEY_WRITE` injection; release binaries do not check the env var.

## Local verification

- `cargo build` — clean (warnings only from upstream `trie-db`).
- `cargo test --release wallet_keys` — 59 passed, 1 ignored.
- `cargo test --release wallet` — 59 passed, 1 ignored.
- PoC test added above passed, demonstrating the HIGH finding, then reverted.

## Verdict

REQUEST CHANGES. The HIGH finding must be fixed before merge — the stated goal of PR #40 ("target is either absent or fully populated, never half-written") is not held on the `--force` path, and any of a dozen common `rename(2)` failure modes can silently destroy the user's coldkey. The MEDIUM and LOW findings can follow in separate PRs.

Do not merge.

moorkh (commented):
[cryptid] round 2 review of 6147c4e

**Verdict: APPROVE** — both round-1 findings sealed, no new findings.

## Build / test parity

Reproduced locally on this worktree:

- `cargo build` — clean
- `cargo clippy --all-targets --all-features -- -D warnings` — clean (only the upstream `trie-db v0.29.1` future-incompat note, not from btt code)
- `cargo test --release wallet_keys` — 62 passed / 0 failed / 1 ignored
- `cargo test --release wallet` — 62 passed / 0 failed / 1 ignored

## HIGH fix (backup-and-rename promote) — verified

Walked `promote_staged_into_existing` step by step (src/commands/wallet_keys.rs:526-676):

1. **Step 1 fails** (`rename(target → backup)`): returns Err before touching staging; target untouched. OK.
2. **Step 1 ok, step 2 fails** (`rename(staging → target)`): rolls back with `rename(backup, target)`. On rollback-success the caller cleans up staging; on rollback-failure the returned `BttError::io` explicitly names the backup path and instructs manual recovery (lines 572-580: *"rollback from {backup} failed ({err}); original wallet may be at the backup path"*). Adequate.
3. **Step 2 ok, step 3 fails** (hotkey merge): new wallet is live, `.bak.*` retains unrelated hotkeys for manual recovery. Error message at line 658 includes both `target_dir` and `backup_dir`. Documented in the function doc-comment.
4. **Step 4 (cleanup backup)**: best-effort `remove_dir_all`; leftover `.bak.*` is harmless forensics. Filtered from `wallet list`.

**Crash window between steps 1 and 2**: yes, `target` is briefly absent. Verified `wallet::list` (src/commands/wallet.rs:29-121) is resilient — it gates on `path.is_dir()` and `coldkey_path.exists()`, so a concurrent read during that window either skips the wallet (if the old backup was seen and filtered) or harmlessly reports it with `coldkey: None`. No crash, no corruption. Transient skip is acceptable.

**Planted-directory PoC from round 1 (EISDIR on coldkeypub.txt)** is no longer reachable: the new publish is a single `rename(2)` on the directory, not a sequence of file-level renames. The round-2 tests intentionally replace that PoC with `BTT_FAIL_BEFORE_PUBLISH` / `BTT_FAIL_DURING_PROMOTE` synthetic hooks, which pin the same invariants at both the pre-publish and post-publish failure points. Test assertions check for the synthetic tag in the error string (`msg.contains(\"BTT_FAIL_BEFORE_PUBLISH\")` etc.) so a wrong-error-path regression would fail the test — not a blanket `is_err()`.

**Hotkey merge error handling**: bails on the first failure via `?`. Remaining hotkeys are not attempted, but they remain in `.bak.*` intact — user has full recovery via the backup. The bail-early behavior is documented in the function doc-comment. Acceptable trade-off; not a blocker.

## MEDIUM fix (force-gate) — verified

src/commands/wallet_keys.rs:463: `if wallet_dir.exists() && force { ... } else if let Err(e) = fs::rename(&staging_dir, &wallet_dir) { ... }`. The no-force branch now falls through to the plain `fs::rename`, which rejects a non-empty target with `ENOTEMPTY` and only succeeds against an empty leftover dir. Pre-round-1 the force path was silently reachable via a crash-leftover dir containing only unrelated hotkeys; that path is closed.

## LOW fix (reserved prefix) — verified

src/commands/wallet_keys.rs:390: prefix check is the very first statement in `create`, ahead of `validate_n_words`. Both `.tmp.` and `.bak.` (with trailing dot) are rejected, so `.tmpfoo` or `.bakfoo` are still allowed — matches the `starts_with(\".tmp.\")` / `starts_with(\".bak.\")` filter in `wallet::list` at src/commands/wallet.rs:68. Symmetric.

## LOW fix (list filter) — verified

src/commands/wallet.rs:68: `if wallet_name.starts_with(\".tmp.\") || wallet_name.starts_with(\".bak.\")`. Covered by `wallet_list_skips_temp_dirs` (extended to plant a `.bak.*` dir, I verified in the diff).

## Nit fixes — verified

10. **Backward-clock collision**: `next_staging_counter()` at src/commands/wallet_keys.rs:698 is a single file-scope `static COUNTER: AtomicU64 = AtomicU64::new(0)` with `fetch_add(1, Ordering::Relaxed)`. Process-local, not per-call. Both `temp_staging_name` and `backup_dir_for` call the same helper, so staging and backup names share the monotonic space — no collision even on a frozen or backward clock. Relaxed ordering is correct for a uniqueness-only counter.

11. **`ensure_secure_dir` on pre-existing target**: confirmed absent from the new promote path. src/commands/wallet_keys.rs:620-628 explicitly comments *"Do NOT call `ensure_secure_dir` here"*. It is only called defensively if `target_hotkeys` is missing post-publish (which shouldn't happen — the staging dir always creates it). User-tightened perms are preserved.

## Failure-injection coverage — verified

12. Both `BTT_FAIL_BEFORE_PUBLISH` (line 561) and `BTT_FAIL_DURING_PROMOTE` (line 599) are wrapped in `#[cfg(test)]`. The production path (line 569-570) is a plain `#[cfg(not(test))] let publish_result = fs::rename(staging_dir, target_dir);` with no env-var access. Release builds have zero overhead and zero surface area.

13. Test assertions check the specific synthetic error tag, not `is_err()`. A wrong-path regression (e.g. permission error instead of injected error) would fail the test.

## Scope and hygiene — verified

14. `git diff --stat ec96a8a HEAD` → only `src/commands/wallet.rs` and `src/commands/wallet_keys.rs`. No scope creep.
15. `Cargo.toml` / `Cargo.lock` unchanged (empty diff). No new deps — consistent with the dependency-discipline rule.
16. All new `.expect(\"...\")` calls in the diff are under `#[cfg(test)]` (verified by scanning the `^+` lines in `git diff`). No new unwraps/expects in production code.
17. Issues [#41](https://github.com/ErgodicLabs/btt/issues/41) (per-wallet flock for concurrent `--force` races) and [#42](https://github.com/ErgodicLabs/btt/issues/42) (reap stale `.tmp.*`/`.bak.*` dirs) are both open. Deferred scope is tracked.

## Remaining failure modes

Thought about concurrent processes, filesystem-level errors, disk-full, PID wraparound. Every remaining failure mode leaves the user with EITHER:

- the old wallet intact (step 1 fail, or step 2 fail with successful rollback), OR
- the new wallet intact plus a `.bak.*` dir for manual recovery of unrelated hotkeys (step 2 success, any later failure), OR
- a clearly-flagged compound-error message naming the backup path (step 2 fail AND step 2-rollback fail — rare, needs concurrent fs activity on the target name).

Never both partial. The `.bak.*` forensic residue is a feature, not a leak, and is filtered from `wallet list`. Cross-process races are out of scope by design and are tracked in #41 under a per-wallet flock.

The new algorithm is clean. The surgery held. Both round-1 findings are sealed.

**APPROVE.** Loop terminates. cryptid will not merge from this review (self-PR), leaving the actual merge to the operator.

Issue #42, opened 2026-04-14
[cryptid] wallet cleanup: reap stale .tmp.* and .bak.* directories


moorkh:
## Context

PR #40 landed the atomic `wallet create` path. It writes staged wallets into `<wallets>/.tmp.<name>.<pid>.<nanos>.<ctr>/` and — on `--force` — backs old wallets up to `<wallets>/.bak.<name>.<pid>.<nanos>.<ctr>/` during the swap. Both prefixes are reserved: `wallet list` filters them out of user-facing output, and `wallet create` refuses to accept a wallet name starting with either prefix.

Both dirs are deliberately **not** auto-cleaned on startup. On a successful run they are cleaned up in-process (`.tmp.*` is renamed into place, `.bak.*` is `remove_dir_all`'d after the hotkey merge). On a crash or a rare failure path (e.g. a promote that fails after the publish rename but before the hotkey merge), stale dirs are left on disk for forensics.

Over time, repeated crashed runs will accumulate `.tmp.*` / `.bak.*` directories under `<wallets>/`. They are harmless (filtered from `wallet list`, never decrypted) but they waste disk space and clutter the directory.

## Proposal

Add a `btt wallet cleanup` subcommand:

```
btt wallet cleanup [--dry-run] [--wallet <name>]
```

Behavior:
- With no arguments: list all `.tmp.*` and `.bak.*` dirs under `<wallets>/`, along with their size and mtime.
- With `--dry-run`: print what would be removed, remove nothing.
- Without `--dry-run`: `remove_dir_all` each stale dir. Log each removal to stderr.
- With `--wallet <name>`: restrict the scan to `.tmp.<name>.*` and `.bak.<name>.*` entries.

Safety invariants:
- NEVER touch anything that does not match the exact reserved prefix regex. The prefixes are:
  - `.tmp.<name>.<pid>.<nanos>.<ctr>`
  - `.bak.<name>.<pid>.<nanos>.<ctr>`
  where `<name>` matches the normal wallet name grammar, `<pid>` and `<ctr>` are integers, and `<nanos>` is up to 19 digits.
- A user wallet named `.tmp.foo` cannot exist (`create` refuses the prefix), so pattern-matching is safe.
- Age filter: optional `--older-than <duration>` (e.g. `7d`) to only reap dirs older than N seconds/minutes/hours/days. Defaults to \"all\" because a stale staging dir from a crashed run is safe to remove immediately — the only reason to keep one is forensics.

[...]

PR #43, opened 2026-04-14
[cryptid] wallet create: per-wallet flock(2) for concurrent --force races (#41)

### LINKED ISSUE #41: [cryptid] wallet create: per-wallet flock to serialize concurrent --force races
## Context

PR #40 (round 2) landed a backup-and-rename algorithm for the `wallet create --force` promote path. The new algorithm is dramatically safer than the pre-#40 sequential-file-rename approach, but it does not address the concurrent-`--force` race between two simultaneous `wallet create --force --name foo` invocations.

## The race window

The promote sequence is:

1. `rename(target, .bak.<name>.<pid_A>.<nanos_A>.<ctr_A>)`  — move old target aside
2. `rename(staging_A, target)`  — atomic publish
3. merge unrelated hotkeys from `.bak.*` into `target/hotkeys/`
4. `remove_dir_all(.bak.*)`

Two processes A and B racing on the same wallet name can interleave between steps 1 and 2:

- A: step 1 succeeds — target is now at `.bak.<pid_A>...`, target name is absent.
- B: step 1 — `rename(target, .bak.<pid_B>...)` — fails with `ENOENT` because target does not exist. B surfaces an error and exits.

Or worse:

- A: step 1 succeeds, target name is absent.
- B: step 1 — `rename(target, .bak.<pid_B>...)` — fails with `ENOENT`, B errors out. But by this point A is still between step 1 and step 2, and if A now fails step 2, its rollback (`rename(.bak.<pid_A>, target)`) races against B's error-return — in practice A's rollback will usually succeed because B is not doing any concurrent filesystem ops on target, but the invariant is fragile.

The larger problem: if A completes all four steps while B is in its own step 1, B's first rename can succeed if it fires after A's step 2 but before A's step 4 (A's target dir is A's new wallet by that point). B will then move A's freshly-published wallet into `.bak.<pid_B>...` and publish its own staging dir. A's step 4 then `remove_dir_all`s B's backup, which is A's just-published wallet. Net result: A's wallet is deleted, only B's survives, and there's no error surfaced to A.

## Proposed fix: `flock(2)` on a per-wallet lock file

Before step 1 of the promote path (and before the simple non-force `fs::rename(staging, target)` too), `open()` + `flock(LOCK_EX)` on a per-wallet lock file at `<wallets>/.lock.<name>`. The lock file itself is persistent — creating and keeping it around is cheap, and it gives us a stable handle to `flock` on. Release on scope exit (the File descriptor drop closes the fd, which releases the advisory lock).

`flock(2)` semantics:
- Advisory lock, not mandatory, but both btt processes voluntarily acquire it so that is sufficient.
- Shared across processes on the same host (unlike `fcntl` POSIX locks, which are per-process file-table records).
- Released automatically on process exit if btt crashes while holding it.
- Non-NFS-safe, but btt wallets live under `$XDG_CONFIG_HOME` which is local.

A failed `flock(LOCK_EX | LOCK_NB)` could either:
1. Block (`LOCK_EX` without `NB`) — simplest, but a hung btt would hang all peers.
2. Return an error after a short timeout — prevents hang, requires retry logic.
3. Return an error immediately — cleanest UX (\"another btt process is writing wallet X, try again\"), trivial to implement.

Recommend option 3. The error message can name the lock file so users can forensically diagnose (`lsof .lock.foo` will name the holding process).

## Scope

- Add `flock(2)` call inside `create` before `ensure_wallets_root` (or just before the staging+publish phase).
- Windows equivalent: `LockFileEx` on the same lock file. Or defer Windows support to a follow-up — btt is primarily targeting Linux/macOS today.
- Test: two threads inside a single test process both calling `create` with the same wallet name; assert exactly one succeeds and the other fails with a lock-busy error. (True cross-process testing requires a subprocess harness.)
- Also file in this lock path: new-coldkey, new-hotkey, regen-coldkey, regen-hotkey — any operation that can destroy a wallet under `--force`.

## Severity

Medium. The Round-1/Round-2 atomicity fix in PR #40 eliminates the half-write regression inside a single process; the remaining concurrent-race exposure requires two deliberate simultaneous `--force` invocations against the same wallet name, which is an unusual workflow. Filing as a follow-up so PR #40 can merge without blocking on a larger refactor.

moorkh:
## Summary

Closes #41.

PR #40 round 2 landed a backup-and-rename promote path for `wallet create --force`. It is dramatically safer than the pre-#40 sequential-file algorithm but still carries a concurrent-process race: two simultaneous `btt wallet create --force --name foo` invocations from different processes can interleave their three rename operations between `rename(target, .bak.*)` and `rename(staging, target)` and silently delete one of the two newly-published wallets. The full interleaving analysis is in issue #41.

This PR closes that race by acquiring a per-wallet-name `flock(LOCK_EX)` advisory lock before any filesystem mutation in `create()`.

## Approach

- A per-wallet sentinel file lives at `<wallets>/.lock.<name>`, created 0600 on first use and reused across runs (no auto-cleanup).
- `flock(LOCK_EX)` is acquired at the very top of `create()` — **before** `guard_create_overwrite`, **before** the reserved-prefix check's filesystem side, **before** anything else that touches the wallet directory. The lock is held for the full duration of `create` and released on scope exit when the owning `File` drops (the explicit `LOCK_UN` in `Drop` is belt-and-braces; `close(2)` releases the advisory lock regardless).
- Granularity is per-wallet-name. Two `create --name alice` invocations serialize; `create --name alice` and `create --name bob` run in parallel because they acquire different sentinel files. Locking all of `wallets/` would be overly conservative.

## No new dependencies

The `flock` syscall is declared via a one-symbol `extern \"C\"` block inside a private `flock_sys` module, rather than pulling `libc`, `nix`, or `fs2`:

```rust
#[cfg(unix)]
mod flock_sys {
    extern \"C\" {
        fn flock(fd: i32, op: i32) -> i32;
    }
    const LOCK_EX: i32 = 2;
    const LOCK_UN: i32 = 8;
    // ...
}
```

`flock(2)`, `LOCK_EX`, and `LOCK_UN` have been POSIX-stable on linux, macOS, and BSDs for 30+ years. A direct FFI declaration has a smaller blast radius than a new transitive crate, and it honors the dep-discipline rationale from PR #21 (which removed `libc` in the first place). `Cargo.toml` is untouched.

## Lock acquisition site

`src/commands/wallet_keys.rs`, inside `create()`, immediately after `validate_n_words` and before `guard_create_overwrite`:

```rust
ensure_wallets_root()?;
let lock_path = paths::wallets_dir()?.join(format!(\".lock.{wallet_name}\"));
let mut lock_open = fs::OpenOptions::new();
lock_open.read(true).write(true).create(true).truncate(false);
#[cfg(unix)]
{
    lock_open.mode(0o600);
}
let lock_file = lock_open.open(&lock_path)?;
let _lock = flock_sys::LockGuard::acquire(lock_file)?;
// ... guard_create_overwrite, staging, promote run while _lock is held ...
```

The lock drops when `create()` returns, on both the success and error paths.

## Reserved-prefix set

Expanded from `.tmp.` / `.bak.` to also include `.lock.`:

- `wallet_keys::create` refuses wallet names with any of the three prefixes (fail fast with `InvalidInput`).
- `wallet::list` skips entries with any of the three prefixes. The existing `!path.is_dir()` check already filters lock files since they are zero-byte regular files; the explicit `.lock.*` skip is defense in depth.

## Tests added (5 new, all `#[cfg(unix)]`, all hold `HOME_LOCK`)

- `wallet_create_acquires_lock` — background thread grabs the lock and holds for 200ms; main thread calls `create` and asserts elapsed wall time >= 100ms, proving the call actually blocked on the flock.
- `wallet_create_releases_lock_on_success` — after a successful `create`, a second thread re-acquires the lock within a 2s timeout. A leaked lock would hang indefinitely.
- `wallet_create_releases_lock_on_error` — triggers a promote-path error via `BTT_FAIL_BEFORE_PUBLISH` and confirms the lock is still released (re-acquired within 2s from a second thread).
- `wallet_create_rejects_reserved_prefix` — extended to also reject `.lock.foo`, asserting the error is `InvalidInput` and mentions the `.lock.` prefix.
- `wallet_list_skips_lock_files` — plants `.lock.example` in `<wallets>/`, runs `wallet::list`, asserts the lock file is not surfaced.

All tests use `std::thread::scope` / `std::sync::Barrier` / `std::sync::mpsc` for synchronization — no new deps.

## Windows fallback

Windows gets a no-op `LockGuard` stub under `#[cfg(not(unix))]`. Concurrent `wallet create --force` on windows is **undefined behavior** — the race documented in issue #41 is left unaddressed on windows pending a follow-up that uses `LockFileEx` on the same sentinel file. Linux and macOS are the primary btt targets, and the round-1 reviewer of #41 accepted that windows support can land later.

## Test plan

- [x] `cargo build` — passes
- [x] `cargo clippy --all-targets --all-features -- -D warnings` — passes (the new `unsafe` block is clippy-clean)
- [x] `cargo test --release wallet_keys` — 66 passed, 0 failed, 1 ignored
- [x] `cargo test --release wallet` — 66 passed, 0 failed, 1 ignored
- [ ] Manual two-process smoke test (run two `btt wallet create --force --name foo` invocations from separate shells, confirm they serialize) — worth doing before merge.

Cargo.toml is untouched. No new deps.

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

### DIFF
--- a/src/commands/wallet.rs
+++ b/src/commands/wallet.rs
@@ -55,17 +55,22 @@ pub fn list() -> Result<WalletList, BttError> {
             .to_string_lossy()
             .to_string();
 
-        // Skip staging and backup directories left behind by
-        // `wallet create` (issue #29). The atomic-create path writes a
-        // sibling `.tmp.<wallet>.<pid>.<nanos>.<ctr>` staging dir, and
+        // Skip staging, backup, and lock files left behind by
+        // `wallet create`. The atomic-create path writes a sibling
+        // `.tmp.<wallet>.<pid>.<nanos>.<ctr>` staging dir (issue #29),
         // the `--force` promote path writes a sibling
         // `.bak.<wallet>.<pid>.<nanos>.<ctr>` backup dir during the
-        // swap. Both prefixes are reserved (see `wallet_keys::create`,
-        // which refuses to create a wallet with either prefix). Stale
-        // `.tmp.*` / `.bak.*` dirs from a crashed run are left on disk
-        // deliberately (forensics) but must not appear in `wallet list`
-        // as if they were real wallets.
-        if wallet_name.starts_with(".tmp.") || wallet_name.starts_with(".bak.") {
+        // swap (PR #40), and the per-wallet `flock(2)` sentinel lives
+        // at `.lock.<wallet>` (issue #41). All three prefixes are
+        // reserved — `wallet_keys::create` refuses to create a wallet
+        // whose name starts with any of them. Stale `.tmp.*` / `.bak.*`
+        // dirs from a crashed run and persistent `.lock.*` sentinels
+        // are deliberately left on disk but must not appear in `wallet
+        // list` as if they were real wallets.
+        if wallet_name.starts_with(".tmp.")
+            || wallet_name.starts_with(".bak.")
+            || wallet_name.starts_with(".lock.")
+        {
             continue;
         }
 
--- a/src/commands/wallet_keys.rs
+++ b/src/commands/wallet_keys.rs
@@ -6,6 +6,100 @@ use std::os::unix::fs::{OpenOptionsExt, PermissionsExt};
 use std::path::{Path, PathBuf};
 use std::sync::atomic::{AtomicU64, Ordering};
 
+/// Per-wallet `flock(2)` advisory locking for the `wallet create` path.
+///
+/// Two simultaneous `btt wallet create --force --name foo` invocations
+/// from different processes could interleave the backup-and-rename
+/// sequence in [`promote_staged_into_existing`] and lose the user's
+/// wallet (see issue #41 for the full race analysis). The fix is a
+/// per-wallet-name lock file at `<wallets>/.lock.<name>` held under
+/// `flock(LOCK_EX)` for the duration of `create`, so concurrent creates
+/// on the same name serialize while creates on different names remain
+/// independent.
+///
+/// We declare `flock` directly via an `extern "C"` block rather than
+/// pulling `libc`, `nix`, or `fs2` as a new dependency — see PR #21 for
+/// the dep-discipline rationale. `flock(2)`, `LOCK_EX`, and `LOCK_UN`
+/// are POSIX and their ABI has been stable on linux, macOS, and BSDs
+/// for decades; a one-symbol FFI declaration is safer and smaller than
+/// the blast radius of a new transitive crate.
+#[cfg(unix)]
+mod flock_sys {
+    use std::fs::File;
+    use std::io;
+    use std::os::unix::io::AsRawFd;
+
+    // POSIX advisory lock constants from <sys/file.h>. Stable ABI on
+    // linux, macOS, and BSDs — these values have not changed in 30+
+    // years. We declare `flock()` directly to avoid pulling `libc` as
+    // a new dep; see PR #21.
+    extern "C" {
+        fn flock(fd: i32, op: i32) -> i32;
+    }
+    const LOCK_EX: i32 = 2;
+    const LOCK_UN: i32 = 8;
+
+    /// Owns a [`File`] handle with a `LOCK_EX` advisory lock held on
+    /// its file descriptor. The lock is released on drop — explicitly
+    /// via `LOCK_UN` for clarity, and implicitly by `close(2)` when
+    /// `File` drops immediately afterwards (a redundancy that is also
+    /// a safety net if the explicit call ever fails).
+    pub struct LockGuard {
+        // Kept alive until drop. The file is never read or written —
+        // it exists only as a stable fd to carry the flock.
+        _file: File,
+    }
+
+    impl LockGuard {
+        /// Acquire `LOCK_EX` on `file`'s descriptor. Blocks until the
+        /// lock is available. Returns an I/O error if the syscall
+        /// fails.
+        pub fn acquire(file: File) -> io::Result<LockGuard> {
+            let fd = file.as_raw_fd();
+            // SAFETY: `fd` is a valid, open file descriptor owned by
+            // `file` for the duration of this call. `flock` with
+            // `LOCK_EX` has no memory-safety preconditions beyond a
+            // valid fd, and we check the return value.
+            let rc = unsafe { flock(fd, LOCK_EX) };
+            if rc != 0 {
+                return Err(io::Error::last_os_error());
+            }
+            Ok(LockGuard { _file: file })
+        }
+    }
+
+    impl Drop for LockGuard {
+        fn drop(&mut self) {
+            let fd = self._file.as_raw_fd();
+            // SAFETY: `fd` is still a valid fd (the `File` has not
+            // been dropped yet; it drops immediately after this call
+            // returns). `flock` with `LOCK_UN` has no memory-safety
+            // preconditions. We ignore the return value — `close(2)`
+            // on the file in the `File` drop will release the lock
+            // regardless.
+            let _ = unsafe { flock(fd, LOCK_UN) };
+        }
+    }
+}
+
+#[cfg(not(unix))]
+mod flock_sys {
+    use std::fs::File;
+    use std::io;
+
+    /// Windows stub: we do not currently implement file locking on
+    /// windows. Concurrent `wallet create --force` on windows is
+    /// undefined behavior; the race documented in issue #41 is left
+    /// unaddressed pending a follow-up that uses `LockFileEx` on the
+    /// same sentinel file. Linux/macOS are the primary btt targets.
+    pub struct LockGuard;
+    impl LockGuard {
+        pub fn acquire(_file: File) -> io::Result<LockGuard> {
+            Ok(LockGuard)
+        }
+    }
+}
+
 use serde::{Deserialize, Serialize};
 use sp_core::crypto::{Pair as TraitPair, Ss58Codec};
 use sp_core::sr25519::{Pair, Public, Signature};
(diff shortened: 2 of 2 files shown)


moorkh (commented):
[cryptid] barbarian review — APPROVE (with two non-blocking notes)

Hostile review of the unsafe FFI + flock(2) serialization for `wallet create`.

## Verification performed (in worktree)

- `cargo build`: clean.
- `cargo clippy --all-targets --all-features -- -D warnings`: clean. The `unsafe` blocks in `flock_sys` are clippy-silent.
- `cargo test --release wallet_keys`: 66 passed, 0 failed, 1 ignored, 250.31s.
- `cargo test --release wallet`: 66 passed, 0 failed, 1 ignored, 246.27s.
- `git diff --stat HEAD~1 HEAD`: only `src/commands/wallet.rs` (+17/-12) and `src/commands/wallet_keys.rs` (+372/-14). `Cargo.toml` / `Cargo.lock` untouched — dep discipline preserved (PR #21).
- `nm -D target/debug/btt | grep flock` → `U flock@GLIBC_2.2.5`. The one-symbol `extern \"C\"` resolves against the implicitly-linked `libc.so.6`. `ldd` confirms `libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6`. Symbol has been stable since glibc 2.2.5 (2002).

## Unsafe FFI — constants and soundness

**LOCK_EX=2, LOCK_UN=8 are correct.** Verified against `/usr/include/x86_64-linux-gnu/bits/fcntl-linux.h:237-241`:

```
# define LOCK_SH  1   /* Shared lock.  */
# define LOCK_EX  2   /* Exclusive lock.  */
# define LOCK_NB  4   /* Or'd with one of the above to prevent ... */
# define LOCK_UN  8   /* Remove lock.  */
```

These values are inherited unchanged from 4.2BSD's original `flock(2)` and are identical on linux, macOS (xnu `bsd/sys/fcntl.h`), FreeBSD, OpenBSD, and NetBSD. They have been stable for 40+ years. `#[cfg(unix)]` is the correct gate.

**`extern \"C\" fn flock(fd: i32, op: i32) -> i32`**: `RawFd` is `i32` on every supported Unix target triple rustc emits (linux, darwin, all BSDs). Signature matches the C prototype `int flock(int, int)`. Sound.

**`unsafe` block minimality**: each call is wrapped in exactly one `unsafe { flock(...) }` expression — the smallest possible scope. Acquire path checks the return value and converts to `io::Error::last_os_error()`; drop path uses `let _ =` to swallow errors (drop must not panic — correct).

**fd lifetime**: `LockGuard` owns the `File` by value (`_file: File`). `as_raw_fd()` borrows; the fd is valid for the full lifetime of the guard. On drop, the explicit `LOCK_UN` fires first, then the `File` drop closes the fd and the kernel releases any residual lock state. The double release is harmless: the second release (via close) on an already-unlocked fd is a no-op; flock on a closed fd would EBADF, but by construction the order is `unlock → close`, not `close → unlock`. Sound.

**`mem::forget(guard)` risk**: would leak the file and leave the lock held indefinitely on that fd. This is memory-safe (no UB) — just a liveness issue. Not reachable from any call site. Acceptable.

## Lock semantics

- Open mode: `.read(true).write(true).create(true).truncate(false)` + `OpenOptionsExt::mode(0o600)` gated on `#[cfg(unix)]`. Correct — NOT `create_new(true)`, so the file is reused across runs as documented. 0600 perms match the rest of the wallet material.
- Call order (lines 485-547): reserved-prefix reject → `validate_n_words` → `ensure_wallets_root` → open lock file → `LockGuard::acquire` → `guard_create_overwrite` → stage → publish → drop guard at function return. `ensure_wallets_root()` runs BEFORE lock open, so a missing `<wallets>/` directory surfaces with a clean error instead of \"failed to open wallet lock file\". (Note: `ensure_wallets_root()` is called a second time at line 557 before re-reading `wallets_dir()` — harmless double-call, minor cleanup opportunity.)
- Per-wallet granularity: `.lock.{wallet_name}` gives one sentinel per name. `alice` and `bob` proceed in parallel; two `alice` serialize. Correct.

## Test quality

- **`wallet_create_acquires_lock`**: bg thread calls `acquire_wallet_lock(wallet_name)`, which opens the same `<wallets>/.lock.<wallet_name>` path with the same `OpenOptions` and calls into the same `flock_sys::LockGuard::acquire` — i.e. it exercises the production locking path, not a parallel implementation. Test would fail if `create()` didn't actually take the lock. `>=100ms` slack on a 200ms bg hold is generous but defensible; tightening to 150ms would be stricter without being flaky, but not a blocker.
- **`wallet_create_releases_lock_on_success`** / **`..._on_error`**: release is verified by a SECOND THREAD acquiring the lock under a `recv_timeout(2s)`. This is a real test — if the guard leaked, the channel recv would time out and the test would fail with \"lock must be free after ... create\". Not window dressing. The error test also uses `BTT_FAIL_BEFORE_PUBLISH` on the `--force` path to drive an actual error through the guard's drop, and the prior baseline `create()` is necessary to populate the target so the force path triggers.
- **`wallet_create_rejects_reserved_prefix`**: extended to `.lock.foo`, asserts `ErrorCode::InvalidInput`, asserts error message mentions both \"reserved prefix\" and \".lock.\". Solid.
- **`wallet_list_skips_lock_files`**: plants `.lock.example` as a zero-byte file, asserts `wallet list` filters it. Note: this test actually exercises the `is_dir()` filter path (the planted file is zero bytes, not a dir), NOT the explicit `.lock.*` starts_with filter. The explicit skip would only fire if someone `mkdir .lock.example` — that case isn't tested. Still correct (either filter catches it), but the test name overpromises.
- All 5 new tests take `HOME_LOCK` and use `seat_home`/`restore_home`. Thread scoping is `std::thread::scope` + `Barrier` in the acquire test (clean hand-off, no sleeps before the timing check); `mpsc::channel` + `recv_timeout` in the release tests (clean bounded-wait).

## Reserved prefix and list filter

- `wallet.rs:49` still has the `is_dir()` filter (first line of defense against zero-byte `.lock.*` sentinels). `wallet.rs:70-74` adds the explicit `.lock.` skip alongside `.tmp.` / `.bak.`. Both present — defense in depth as claimed.
- Trailing-dot consistency: `wallet_name.starts_with(\".lock.\")` requires the trailing dot, matching `.tmp.` / `.bak.` behavior. `.lockfoo` (no dot) is NOT rejected. Consistent.

## Non-blocking notes (file as follow-ups if desired)

**NOTE-1 (LOW) — manual lock-file unlink race.** If a user (or a cleanup script) `rm`s `<wallets>/.lock.<name>` while a `create` is mid-execution, the holder's fd still points to the old inode and keeps the kernel lock entry on that inode. A concurrent second `create` would `open(... create(true))` a NEW inode at the same path, acquire a LOCK_EX on that new inode, and proceed in parallel with the first holder — defeating the serialization. This is a real race, but the exploit surface is narrow: it requires a human (or a misbehaving script) actively rm'ing a dotfile inside `<wallets>/` during a wallet create. Mitigations for a follow-up: (a) document the invariant that `<wallets>/.lock.*` must not be manually removed; (b) `fstat` the fd after acquire and re-`stat` the path, bail if the inodes don't match; (c) move the lock to the `<wallets>/` directory fd itself at the cost of coarser granularity. I would not block merge on this — flag it in issue #42 (the reap-stale-sentinels work already referenced in the PR comment).

**NOTE-2 (INFO) — `wallet_list_skips_lock_files` test doesn't exercise the explicit skip.** The planted `.lock.example` is a zero-byte file; `is_dir()` filters it before the `.lock.*` starts_with check runs. If the explicit skip regressed, this test would still pass. A stricter variant would also `fs::create_dir_all(wallets_parent.join(\".lock.dirtest\"))` and assert it's skipped — that would exercise the explicit filter. Minor.

**NOTE-3 (INFO) — double `ensure_wallets_root()`.** Called at line 515 (before lock open) and again at line 557 (before staging). The second call is redundant but harmless. Cleanup opportunity, not a defect.

## Verdict

**APPROVE**. The unsafe FFI is minimal, correct, and sound. The constants match the system headers. The symbol resolves cleanly against the implicitly-linked libc. The lock is held across the exact check-then-write window that issue #41 identified, with per-wallet granularity and correct drop semantics. Tests exercise production code paths (not parallel implementations) and verify release via second-thread acquisition under bounded timeout. Scope is clean (two files, no Cargo changes). Dep discipline preserved. Do not merge (self-PR, blocks --approve)."

PR #44, opened 2026-04-14
[cryptid] wallet_keys: batched hygiene cleanups (#9)

### LINKED ISSUE #9: wallet_keys hygiene cleanups: fsync, lock poison, perms, dead code
[cryptid]

Batched follow-up issue for the LOW and INFO findings from the PR #3 round-2 barbaric review. None individually warrant their own issue but together they're worth a small cleanup PR.

### NEW-L1: no parent-directory fsync in `write_secure_file`
`src/commands/wallet_keys.rs`, the secure write helper. After `sync_all()` on the file, the directory entry that points at the file is not fsynced. On a hard crash between the file write and the directory write, the file may exist with content but be invisible (or vice versa). Standard durability fix: also `File::open(parent_dir)?.sync_all()?` after writing the file.

### NEW-L2: `HOME_LOCK.lock().expect(...)` panics on poisoning
The test-side mutex serializing HOME-mutating tests. If a test panics while holding the lock, the lock becomes poisoned and every subsequent test that grabs it panics with "PoisonError". The standard pattern is `.unwrap_or_else(|e| e.into_inner())` so a poisoned lock continues to function. Doesn't affect production code, only test reliability.

### NEW-L3: `coldkeypub.txt` uses umask-default perms
The public-key file is written with the default umask (typically 0644 or 0664), not explicitly set to 0644. Cosmetic — the contents are public so the perms don't matter — but inconsistent with the explicit-perms posture of the rest of the wallet directory. Set explicitly to 0644 for clarity.

### NEW-I1: dead branch in `decrypt_coldkey_interactive`
The function still has an unreachable branch from a prior refactor. Compiler won't warn because it's reachable through one code path that no caller uses. Remove.

### NEW-I2: 24-word generation hard-rejected
`generate_with_phrase` from sp-core supports both 12 and 24 words via the underlying substrate-bip39, but the btt CLI hard-rejects 24-word requests with an error. Mentioned as a deferred functional gap in PR #3's commit message — tracked here for the eventual fix. Not a bug, just an unfilled feature.

## Severity

LOW (NEW-L1, L2, L3) and INFO (NEW-I1, I2). None block anything. Combine into one cleanup PR after the higher-priority queue is clear.

## Related

- Originating findings: PR #3 round-2 review https://github.com/ErgodicLabs/btt/pull/3#issuecomment-4236598774

moorkh:
[cryptid]

Closes #9.

Four low/info follow-ups from the PR #3 round-2 barbarian review, batched into one small cleanup PR per the issue. All scoped to `src/commands/wallet_keys.rs`.

### NEW-L1 — parent-directory fsync in `write_secure_file`

`src/commands/wallet_keys.rs:1408` (`write_secure_file`) and `:1443` (new `write_public_file`) both call a new `sync_parent_dir()` helper at `:1478` after the file's own `sync_all()`. On a hard crash between the file inode checkpoint and the directory dirent checkpoint, ext4/xfs can leave the written file either orphaned (content on disk, name not visible) or invisible (name gone, content still linked). Opening the parent as a `File` and calling `sync_all()` on it flushes the dirent. The helper is `cfg(unix)`-gated — opening a directory as `File` is legal on unix but errors with `ERROR_ACCESS_DENIED` on windows, and the finding is unix-specific.

### NEW-L2 — `HOME_LOCK` poison recovery (test-only)

The 35 `HOME_LOCK.lock().expect("home lock")` sites in the `wallet_keys` test module (starts at `src/commands/wallet_keys.rs:1574`, callers throughout `:1803-3267`) are replaced with `.unwrap_or_else(|e| e.into_inner())`. If a test panics while holding the mutex the lock becomes poisoned, and every subsequent HOME-mutating test was panicking with `PoisonError` instead of proceeding. Test-only change; production code is unaffected.

### NEW-L3 — `coldkeypub.txt` explicit 0644

The three `fs::write(.../coldkeypub.txt, ...)` sites — `stage_wallet` at `src/commands/wallet_keys.rs:904`, `new_coldkey` at `:998`, `regen_coldkey` at `:1101` — now call a new `write_public_file()` helper (`:1443`) that opens with `OpenOptionsExt::mode(0o644)` on unix, mirroring the `write_secure_file` posture for secret files. Contents are public so the bit that matters is consistency with the rest of the wallet dir's explicit-perms posture; previously the file landed at umask-default (typically 0664 on a user shell, 0666 under a loose umask).

### NEW-I1 — dead unencrypted branch in `decrypt_coldkey_interactive`

The "try unencrypted JSON first" branch at `src/commands/wallet_keys.rs:201-208` is deleted. Every btt-produced coldkey is a binary `$NACL` envelope written by `encrypt_key_data` + `write_secure_file`, so the UTF-8 check on a real coldkey fails at the first non-ASCII byte of the nonce and the branch cannot be reached by any production caller (only `stake add` / `stake remove` call this function, and both always see encrypted coldkeys). Leftover from a pre-encryption prototype. Tests that need an unencrypted round-trip go through the private `load_coldkey` / `pair_from_key_json` helpers directly, which already exist.

## Tests added (3)

All `cfg(test)` inside the existing `wallet_keys` tests module at the bottom of the file.

- **`write_secure_file_success_path_with_parent_fsync`** — pins the happy path of `write_secure_file` + `sync_parent_dir`. Writes a payload, reads it back, asserts 0600 mode, then re-writes the same path to exercise the remove-then-create-new branch plus a second parent-dir fsync.
- **`home_lock_recovers_from_poisoning`** — spawns a sub-thread that panics while holding `HOME_LOCK`, asserts the subsequent `HOME_LOCK.lock()` returns `Err(PoisonError)`, and confirms `.unwrap_or_else(|e| e.into_inner())` yields a usable guard. Pins the NEW-L2 pattern so a regression back to `.expect("home lock")` would fail the suite.
- **`coldkeypub_txt_mode_is_0644`** — sets `umask(0)` process-wide (so a stray `fs::write` would land at 0666), runs `create`, reads `coldkeypub.txt`'s metadata, asserts the mode is exactly `0o644`. `umask(2)` is declared directly as `extern "C"` rather than pulling `libc` — same dep-discipline pattern as the existing `flock_sys` module (see PR #21). `umask` is restored before assertions so a failed test cannot leak the relaxed mask to sibling tests.

### Why NEW-L1 has no direct durability test

Testing actual parent-dir-fsync durability requires crash injection — something that can kill the process between the file fsync and the parent fsync and then observe whether the dirent is on stable storage. btt has no such harness, and introducing one for a LOW finding would be disproportionate. The happy-path test pins that `sync_parent_dir` does not break the main write routine (wrong parent, permission error on dir open, missing parent handling); durability itself is covered by the code review of the syscall pair.

### Why NEW-I1 has no dedicated test

The branch was genuinely dead. The existing 66 `wallet_keys` tests exercise the encrypted path through `decrypt_coldkey_interactive` indirectly (via `sign`, `stake`) and all pass after deletion.

## Verification

- `cargo build`: ok
- `cargo clippy --all-targets --all-features -- -D warnings`: ok
- `cargo test --release wallet_keys`: 69 passed, 0 failed, 1 ignored (the `dump_blob_for_python` helper)

## No new dependencies

`cargo tree` is unchanged from main. The `umask(2)` call in the new test is declared directly as `extern "C"`, matching the existing `flock_sys` pattern at `src/commands/wallet_keys.rs:27-83` — no `libc` / `nix` / `fs2` pulled in. No touch to `flock_sys` or `promote_staged_into_existing` (PR #43 / #40 territory).

### DIFF
--- a/src/commands/wallet_keys.rs
+++ b/src/commands/wallet_keys.rs
@@ -188,26 +188,14 @@ pub fn decrypt_coldkey_interactive(wallet_name: &str) -> Result<Pair, BttError> 
         )));
     }
 
-    // Try reading as unencrypted JSON first (development/testing keys).
-    // The plaintext contents still hold a mnemonic, so funnel it through
-    // a `Zeroizing<Vec<u8>>` instead of `fs::read_to_string`, which would
-    // leak the bytes to an unscrubbed `String` allocation.
-    //
-    // We skip the old `serde_json::Value` pre-check (which itself would
-    // allocate unscrubbed owned `String`s for every JSON string node) and
-    // go straight to `pair_from_key_json`, which routes the data through
-    // the zeroizing `LoadedKeyFile`. On an encrypted `$NACL` envelope the
-    // UTF-8 check fails and we fall through to the decrypt branch.
-    if let Ok(raw) = fs::read(&coldkey_path) {
-        let raw: Zeroizing<Vec<u8>> = Zeroizing::new(raw);
-        if let Ok(content) = std::str::from_utf8(&raw) {
-            if let Ok(pair) = pair_from_key_json(content) {
-                return Ok(pair);
-            }
-        }
-    }
-
-    // Encrypted: prompt for password and decrypt
+    // Every btt-produced coldkey on disk is a binary `$NACL` envelope written
+    // by `encrypt_key_data` + `write_secure_file`, so the older
+    // "try unencrypted JSON first" branch cannot reach any real caller: the
+    // UTF-8 check on a `$NACL` blob fails at the first non-ASCII byte of the
+    // nonce. The dead branch was a leftover from a pre-encryption prototype
+    // and is deleted per NEW-I1 of the PR #3 round-2 review (issue #9).
+    // Testing keys that need an unencrypted round-trip go through the private
+    // `load_coldkey` / `pair_from_key_json` helpers directly.
     let password = read_password("Enter coldkey password: ")?;
     let pair = load_coldkey(&wdir, &password)?;
     Ok(pair)
@@ -901,8 +889,7 @@ fn create_into_staging(
     let pub_data = build_pub_key_json(&cold_pair);
     let pub_json_str = serde_json::to_string(&pub_data)
         .map_err(|e| BttError::io(format!("failed to serialize coldkeypub: {}", e)))?;
-    fs::write(staging_dir.join("coldkeypub.txt"), &pub_json_str)
-        .map_err(|e| BttError::io(format!("failed to write coldkeypub.txt: {}", e)))?;
+    write_public_file(&staging_dir.join("coldkeypub.txt"), pub_json_str.as_bytes())?;
 
     // Generate hotkey
     let (hot_pair, mut hot_phrase, mut hot_seed) = generate_keypair(n_words)?;
@@ -995,8 +982,7 @@ pub fn new_coldkey(
     let pub_data = build_pub_key_json(&pair);
     let pub_json_str = serde_json::to_string(&pub_data)
         .map_err(|e| BttError::io(format!("failed to serialize coldkeypub: {}", e)))?;
-    fs::write(wallet_dir.join("coldkeypub.txt"), &pub_json_str)
-        .map_err(|e| BttError::io(format!("failed to write coldkeypub.txt: {}", e)))?;
+    write_public_file(&wallet_dir.join("coldkeypub.txt"), pub_json_str.as_bytes())?;
 
     let mnemonic_out = phrase.clone();
 
@@ -1098,8 +1084,7 @@ pub fn regen_coldkey(
     let pub_data = build_pub_key_json(&pair);
     let pub_json_str = serde_json::to_string(&pub_data)
         .map_err(|e| BttError::io(format!("failed to serialize coldkeypub: {}", e)))?;
-    fs::write(wallet_dir.join("coldkeypub.txt"), &pub_json_str)
-        .map_err(|e| BttError::io(format!("failed to write coldkeypub.txt: {}", e)))?;
+    write_public_file(&wallet_dir.join("coldkeypub.txt"), pub_json_str.as_bytes())?;
 
     Ok(RegenResult {
         wallet_name: wallet_name.to_string(),
@@ -1405,6 +1390,13 @@ fn decrypt_key_data(encrypted: &[u8], password: &str) -> Result<Zeroizing<Vec<u8
 /// write `data`, fsync, and close. Fails if the file already exists unless the
 /// existing file can be removed first (callers that intend to overwrite should
 /// delete first). No TOCTOU window at 0644.
+///
+/// After the file's own `sync_all()` returns, the parent directory is also
+/// opened and `sync_all()`d so the new dirent reaches stable storage. Without
+/// this second fsync a power loss between the file's inode checkpoint and
+/// the directory's dirent checkpoint can leave the file either orphaned
+/// (content on disk, name not visible) or invisible (name gone, content
+/// still linked). NEW-L1 from the PR #3 round-2 review (issue #9).
 fn write_secure_file(path: &Path, data: &[u8]) -> Result<(), BttError> {
     // If the file exists, remove it first; we want to replace atomically.
     if path.exists() {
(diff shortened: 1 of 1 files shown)


moorkh (commented):
[cryptid] Hostile review of PR #44 — wallet_keys hygiene cleanups (issue #9)

## Verdict: APPROVE

All four items (NEW-L1/L2/L3/I1) land cleanly. Build, clippy `-D warnings`, and `cargo test --release wallet_keys` (69 passed, 0 failed, 1 ignored) all green locally. The three new hygiene tests are present and passing:
- `write_secure_file_success_path_with_parent_fsync ... ok`
- `home_lock_recovers_from_poisoning ... ok`
- `coldkeypub_txt_mode_is_0644 ... ok`

## NEW-L1 — parent fsync

- `write_secure_file` calls `sync_parent_dir(path)?` at L1420; `write_public_file` at L1454. Both on the success tail.
- Error paths (remove / open / write / fsync failure) do **not** fsync the parent. This is correct POSIX semantics: `Err` means the caller cannot assume durability. Not a regression.
- `sync_parent_dir` opens the parent via `fs::File::open(parent)` (L1473). Linux returns a read-only fd, and `fsync(2)` on a dir fd only requires a valid fd, not write access. Correct.
- Windows: `#[cfg(unix)]`-gated with an explicit doc comment (L1467–L1470). Documented.
- Empty-parent case: handled by `let Some(parent) = path.parent() else { return Ok(()); };` (L1463). No panic.

## NEW-L2 — poison recovery

- 35 call sites of `HOME_LOCK.lock().unwrap_or_else(|e| e.into_inner())`. Zero `.expect(\"home lock\")` remaining in code (one surviving occurrence at L3376 is inside a doc comment describing the old pattern). Full conversion.
- `home_lock_recovers_from_poisoning` test asserts both arms: `poisoned.is_err()` at L3379 (the lock is *actually* poisoned) and `unwrap_or_else(|e| e.into_inner())` at L3384 (recovery works).
- Test does not itself hold `HOME_LOCK` on the main thread, but the sub-thread's `lock()` blocks behind any concurrent HOME-mutating test rather than racing. Every `HOME_LOCK` consumer uses the recovery idiom, so a poisoned lock is benign. Acceptable.

## NEW-L3 — explicit 0644

- `write_public_file` (L1435–L1456): `OpenOptions::new().write(true).create_new(true)` + `#[cfg(unix)] { opts.mode(0o644); }`. No explicit `truncate(true)`, but the file is pre-removed at L1436–L1439, so `create_new` opens a fresh inode every time — the \"stale trailing bytes\" concern doesn't apply.
- Calls `sync_parent_dir(path)?` at L1454 — symmetric with `write_secure_file`. Good.
- `coldkeypub_txt_mode_is_0644` test (L3395) acquires `HOME_LOCK` at L3396 *before* touching process-wide umask. Umask restored at L3416 before the result `.expect(...)` at L3418 so a `create` failure unwinds with the umask already reset.
- `extern \"C\" { fn umask(mask: u32) -> u32; }` at L3404–L3406: on Linux `mode_t` is `unsigned int` (u32), so the declaration matches glibc. On macOS `mode_t` is `u16`, so this would be a size mismatch under the SysV ABI, but CI is Linux and the `SAFETY:` comment at L3401 explains the no-libc-dep rationale (dep discipline PR #21). Minor portability caveat, not a blocker.

**Minor hygiene note (not blocking):** if `seat_home(...)` at L3409 panics between `umask(0)` at L3407 and `umask(old_mask)` at L3416, umask=0 leaks to sibling tests. `HOME_LOCK` serializes *HOME-mutating* tests but umask is process-global, so a parallel non-HOME test could theoretically see the loosened mask. In practice a panic here means the test is already failing and the lock poison will surface the problem; wrapping the seat/create in a drop-guard that restores umask would harden this further but is out of scope for this PR.

## NEW-I1 — dead branch removal

- `grep -rn decrypt_coldkey_interactive src/` returns exactly two callers: `stake.rs:328` (stake add) and `stake.rs:387` (stake remove). Matches PR claim.
- Deleted branch was a UTF-8 → `pair_from_key_json` probe on the raw bytes. Every btt-produced coldkey is a `$NACL`-magic envelope, which fails UTF-8 at the first non-ASCII byte of the 24-byte nonce. The branch was unreachable from either caller. No regression.

## Diff scope

- `git diff --stat HEAD~1 HEAD`: only `src/commands/wallet_keys.rs` touched (+249 / −58).
- `Cargo.toml` / `Cargo.lock` unchanged. No new deps.
- Zero new `.unwrap()` on production paths.

APPROVE. Self-PR blocks `--approve` so this is a comment-review; not merging.

PR #45, opened 2026-04-14
[cryptid] password_file + btcli-compat: workflow path filter + tilde test refactor (#14)

### LINKED ISSUE #14: btcli-compat workflow + password_file: LOW/INFO follow-ups
[cryptid]

Batched LOW and INFO findings from the PR #12 round-1 barbaric review. Each individually is a nit; together they're worth a small cleanup PR after the higher-priority queue clears.

### NEW-L1: workflow paths filter omits `src/main.rs` and `src/cli.rs`
`.github/workflows/btcli-compat.yml` triggers on changes to `src/commands/wallet_keys.rs`, `Cargo.toml`, etc — but a future PR that rewires `--password-file` only via `src/main.rs` (the dispatch helper) or `src/cli.rs` (the flag declaration) would silently skip the verifier. Add `src/main.rs` and `src/cli.rs` to the paths list.

### NEW-L2: `std::env::set_var` in `expand_tilde_prefix` test
Parallel-test footgun (HOME mutation races with other env-touching tests). Will be `unsafe` in Rust 1.82+. Fix: refactor `expand_tilde_prefix` to take `home: Option<&str>` as a parameter, so the test can pass a synthetic value without mutating real env. Avoids needing `serial_test` as a new dep.

### NEW-L3: `fs::metadata` follows symlinks
Compounds the TOCTOU finding (#13) but is not a separate bug: the FIFO check still fires on the symlink target, not the symlink itself. Once #13 is fixed via `O_NOFOLLOW`, this becomes moot.

### NEW-I1: `dtolnay/rust-toolchain@stable` is tag-pinned, not SHA-pinned
De facto standard for rust on Github Actions; the cost of moving to a SHA pin is some maintenance churn for marginal trust improvement. Note for the future. If we ever land the dep-checksum tripwire (issue #6), the same logic should apply to action pins.

### NEW-I2: `shred_and_remove` is symbolic on tmpfs
The script's `shred_and_remove` helper calls `shred -u` on the temp password file, but `shred` on tmpfs / journaled filesystems doesn't actually overwrite the data — the OS's free-block allocator can return the same block on the next write, which may already have been zeroed, but there's no guarantee. Already acknowledged in the docstring; included here for tracking.

## Severity

LOW (NEW-L1, L2, L3) and INFO (NEW-I1, I2). None block anything. Combine into one cleanup PR after the higher-priority queue clears.

## Related

- Originating findings: PR #12 round-1 barbaric review https://github.com/ErgodicLabs/btt/pull/12#issuecomment-4237110231
- Issue #13 — TOCTOU on --password-file open (the MEDIUM finding from the same review)

moorkh:
[cryptid]

Closes #14.

## Summary

Two LOW findings from issue #14, both small. None individually warranted a builder dispatch.

### NEW-L1: workflow path filter omitted \`src/main.rs\` and \`src/cli.rs\`

The btcli-compat verifier triggers on changes to \`wallet_keys.rs\` and \`password_file.rs\` but not on \`src/main.rs\` (the dispatch helper) or \`src/cli.rs\` (the flag declaration). A future PR that rewires \`--password-file\` via either of those without touching the named files would silently skip the verifier. Both files added to the paths list with an inline comment explaining the addition.

### NEW-L2: \`std::env::set_var\` in \`expand_tilde_prefix\` test

The \`expand_tilde()\` helper read \`\$HOME\` via \`std::env::var\`. Its \`expand_tilde_prefix\` test mutated \`\$HOME\` via \`std::env::set_var\`, which is a parallel-test footgun (the env mutation races with any other test that reads \$HOME, and \`std::env::set_var\` becomes \`unsafe\` in Rust 1.82+).

**Refactor**: \`expand_tilde()\` now takes \`home: Option<&str>\` as a parameter. The single production caller (\`read_password_file\`) reads \$HOME once and passes it in. Tests pass synthetic values directly with no env mutation.

### NEW-L3: \`fs::metadata\` follows symlinks

Already moot post PR #21. The file open path uses portable \`std::fs::File::open\` which follows symlinks, then \`file.metadata()\` is fd-based and reflects the opened (target) file. The FIFO check fires on the target, which is the correct semantic. **No code change needed.**

### NEW-I1: \`dtolnay/rust-toolchain\` SHA pin

Future-work note. PR #38 (toolchain pinning) addressed the floating-stable problem with a version pin; SHA pinning is a further improvement nobody is asking for yet. **No code change.**

## Files changed

- \`.github/workflows/btcli-compat.yml\` — paths list + comment, +6 / −0
- \`src/commands/password_file.rs\` — \`expand_tilde\` signature change + caller update + 5 tests, +39 / −9

Net: 2 files, +45 / −9. **No source code behavior change beyond the test surface. No new dependencies.**

## Tests

- \`expand_tilde_absolute_untouched\` updated to take \`None\` and \`Some\` variants (no behavior change)
- \`expand_tilde_relative_untouched\` updated likewise
- \`expand_tilde_prefix\` uses \`Some(\"/home/alice\")\`, no env mutation
- \`expand_tilde_bare\` **new** — covers the bare \`~\` case (previously untested)
- \`expand_tilde_no_home_returns_unchanged\` **new** — covers the \`None\` case for both \`~/foo\` and \`~\`

Result: **22 \`password_file\` tests** (was 20, +2 new), all passing.

## Verification (cryptid host)

- yaml parse on \`btcli-compat.yml\`: ok
- \`cargo build\`: clean
- \`cargo clippy --all-targets --all-features -- -D warnings\`: clean
- \`cargo test --release password_file\`: 22 passed, 0 failed

## Builder context

Hand-written by cryptid (not a builder subagent) because the diff is two small surgical changes — a 6-line YAML addition and a function signature refactor with mechanical caller/test updates.

## Out of scope

- NEW-L3 / NEW-I1 (no-op per above).
- SHA pinning the toolchain (separate decision, low value vs the version pin already in place).

## Related

- PR #21 (the password_file TOCTOU rewrite that made NEW-L3 moot).
- PR #38 (the toolchain version pin).

### DIFF
--- a/.github/workflows/btcli-compat.yml
+++ b/.github/workflows/btcli-compat.yml
@@ -9,6 +9,12 @@ on:
     paths:
       - "src/commands/wallet_keys.rs"
       - "src/commands/password_file.rs"
+      # main.rs and cli.rs added per #14 NEW-L1: a future PR that
+      # rewires --password-file via the dispatch helper or the flag
+      # declaration without touching wallet_keys.rs / password_file.rs
+      # would otherwise silently skip this verifier.
+      - "src/main.rs"
+      - "src/cli.rs"
       - "Cargo.toml"
       - "Cargo.lock"
       - "scripts/btcli-compat/**"
--- a/src/commands/password_file.rs
+++ b/src/commands/password_file.rs
@@ -43,19 +43,25 @@ use zeroize::Zeroizing;
 
 use crate::error::BttError;
 
-/// Resolve `~` prefixes and `~user`-less tilde expansion using `$HOME`.
+/// Resolve `~` prefixes and `~user`-less tilde expansion against the
+/// caller-supplied home directory.
 ///
 /// We deliberately do NOT do arbitrary shell expansion — no `$VAR`
 /// substitution, no glob, no `..` magic. A password file path is supposed to
 /// be a concrete on-disk location.
-fn expand_tilde(path: &str) -> PathBuf {
+///
+/// `home` is taken as a parameter (rather than read from `std::env::var`
+/// internally) so tests can supply a synthetic value without mutating the
+/// process-global env. `std::env::set_var` becomes `unsafe` in Rust 1.82+
+/// and is a parallel-test footgun even before that — see issue #14 NEW-L2.
+fn expand_tilde(path: &str, home: Option<&str>) -> PathBuf {
     if let Some(rest) = path.strip_prefix("~/") {
-        if let Ok(home) = std::env::var("HOME") {
+        if let Some(home) = home {
             return PathBuf::from(home).join(rest);
         }
     }
     if path == "~" {
-        if let Ok(home) = std::env::var("HOME") {
+        if let Some(home) = home {
             return PathBuf::from(home);
         }
     }
@@ -68,7 +74,8 @@ fn expand_tilde(path: &str) -> PathBuf {
 /// other-readable. The first line (stripped of trailing `\n` or `\r\n`) is
 /// returned as the password.
 pub fn read_password_file(path: &str) -> Result<Zeroizing<String>, BttError> {
-    let resolved = expand_tilde(path);
+    let home = std::env::var("HOME").ok();
+    let resolved = expand_tilde(path, home.as_deref());
     read_password_file_inner(&resolved)
 }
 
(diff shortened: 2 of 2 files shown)


moorkh (commented):
[cryptid] **APPROVE**

Hostile review of eee82bd. Build, clippy (`-D warnings`), and `cargo test --release password_file` (22/22) all pass in a fresh worktree. `Cargo.toml`/`Cargo.lock` untouched. Diff scope limited to the two expected files.

**NEW-L1 — workflow path filter.** `src/main.rs` and `src/cli.rs` added under `on.pull_request.paths`, not `push` (which has only `branches`). `python3 yaml.safe_load` confirms the structure. Inline comment correctly cites `#14 NEW-L1` and explains the rationale: a future PR rewiring `--password-file` via the dispatch helper or the flag declaration without touching `wallet_keys.rs` / `password_file.rs` would otherwise silently skip the verifier. Accurate.

**NEW-L2 — `expand_tilde` refactor.** Signature is `fn expand_tilde(path: &str, home: Option<&str>) -> PathBuf`. The function body uses only the parameter — no `std::env::var` inside it. The single production caller `read_password_file` reads `HOME` once:

```rust
let home = std::env::var("HOME").ok();
let resolved = expand_tilde(path, home.as_deref());
```

Lifetime is sound: `home: Option<String>` lives to end of scope; `home.as_deref()` produces `Option<&str>` consumed synchronously by `expand_tilde`, which returns an owned `PathBuf`. No borrow escapes.

**Test module hygiene.** Grepped `set_var` — zero real calls; only appears in the doc comment explaining *why* it was removed (unsafe in 1.82+, parallel-test footgun) and a reassuring comment in `expand_tilde_prefix`. No new `.unwrap()` in production paths.

**Tilde test coverage.** `expand_tilde_bare` covers `~` + `Some("/home/alice")` → `/home/alice`. `expand_tilde_no_home_returns_unchanged` covers BOTH `~/foo` + `None` → `~/foo` AND bare `~` + `None` → `~`. Good.

**End-to-end behavior walkthrough (pre- vs post-refactor):**
- `~/foo` + `HOME=/home/alice` → `/home/alice/foo` — same
- `~/foo` + HOME unset → `~/foo` — same (previously also fell through to `PathBuf::from(path)`)
- `~` + `HOME=/home/alice` → `/home/alice` — same
- `~` + HOME unset → `~` — same
- `foo/bar` / `/abs` + anything → pass-through — same

**Edge case: `HOME=\"\"`.** `env::var` returns `Ok(\"\")`, `.ok()` → `Some(String::new())`, `.as_deref()` → `Some(\"\")`. `expand_tilde(\"~/foo\", Some(\"\"))` → `PathBuf::from(\"\").join(\"foo\")` → `foo`. Same degenerate behavior as before the refactor. Not a regression, acceptable (a user with `HOME=\"\"` has bigger problems).

**Mutation probe.** I mutated `expand_tilde` to ignore the parameter and fall back to `std::env::var(\"HOME\")`, then ran the five tilde tests with `HOME=/nonexistent-probe`. Three tests failed as expected:

- `expand_tilde_prefix`: got `/nonexistent-probe/foo`, wanted `/home/alice/foo`
- `expand_tilde_bare`: got `/nonexistent-probe`, wanted `/home/alice`
- `expand_tilde_no_home_returns_unchanged`: got `/nonexistent-probe/foo`, wanted `~/foo`

The tests actually exercise the parameter path. Mutation reverted; tree clean.

NEW-L3 (moot per #21) and NEW-I1 (future-work SHA pin) correctly declared without code change.

Surgical, complete, no loose ends. Do not merge — self-PR, human approval required.

PR #47, opened 2026-04-14
[cryptid] dep-audit/checksum.py: path+file:// + excludes + exit code + cosmetic (#22)

### LINKED ISSUE #22: dep-checksum tripwire: path+file:// handling and follow-up cleanups
[cryptid]

Batched LOW + INFO findings from the PR #17 round-1 barbaric review: https://github.com/ErgodicLabs/btt/pull/17#issuecomment-4237862413

### LOW-1: `path+file://` URL handling is latently broken
`classify_source` in `scripts/dep-audit/checksum.py` leaves `file:///abs/path` as a non-absolute string, then gets re-joined against `REPO_ROOT`, which hard-fails on any real path dep. btt has zero path deps today so this is cold code, but it will break the moment one lands.

Fix: parse the file:// URL properly and use the absolute path. Also add a test using a synthetic path-dep entry.

### LOW-2: path-dep manifest has no exclusion rules
The deterministic manifest includes `target/`, `.git/`, editor swap files. For any future path dep this is noisy-to-useless — the checksum churns on every build. Need an exclude list (target/, .git/, *.swp, *.swo, *~, .DS_Store).

### INFO-1: exit code 2 promised but never used
The script's docstring promises exit code 2 for script errors, but `raise SystemExit("error: ...")` exits 1. Either fix the docstring or change the raises to `sys.exit(2)`.

### INFO-2: cosmetic — `--update` runs still print 'Run --update' interstitial
The interstitial report still tells the user to run `--update` even when they just did. Cosmetic only.

## Severity

LOW (LOW-1, LOW-2) and INFO (INFO-1, INFO-2). None block anything. Combine into one cleanup PR.

## Related

- PR #17 (the dep checksum tripwire): https://github.com/ErgodicLabs/btt/pull/17
- Round-1 barbarian: https://github.com/ErgodicLabs/btt/pull/17#issuecomment-4237862413

moorkh:
[cryptid]

Closes #22. Batched LOW + INFO follow-ups from the PR #17 round-1 barbaric review of the dep-checksum tripwire. All four fixes are minimum-diff, stdlib-only, no new dependencies.

### LOW-1: `path+file://` URL handling
`scripts/dep-audit/checksum.py:153-176` — `classify_source` now parses `path+file://` sources via `urllib.parse.urlparse` and returns the absolute filesystem path. Previously the `file://` scheme was left in the rest string and re-joined against `REPO_ROOT`, which would produce a nonsense path on any real path dep. btt has zero path deps today so this was cold code; the fix unblocks the moment one lands.

### LOW-2: path-dep manifest exclusions
`scripts/dep-audit/checksum.py:60-67` (new module-level constants) and `scripts/dep-audit/checksum.py:215-227` (the walk in `checksum_path`) — the deterministic file walker now prunes `target/` and `.git/` from `os.walk` and skips `*.swp`, `*.swo`, `*~`, `.DS_Store` via `fnmatch`. The exclusion lists are constants at the top of the file with a comment explaining the purpose. Without this, any future path dep's checksum would churn on every cargo build.

### INFO-1: exit code 2 on script error
`scripts/dep-audit/checksum.py:88-99` (new `_die` helper) and the seven script-error sites at `parse_cargo_lock` (line ~111), `checksum_registry` (line ~177), `checksum_git` (lines ~191, ~196), `checksum_path` (lines ~220, ~239, ~250), `compute_entry` (line ~270), and `load_db` (line ~288) — all `raise SystemExit(f"error: ...")` calls (which exited 1) are replaced with `_die(...)` which prints to stderr and raises `SystemExit(2)`, matching the docstring contract (1 = mismatch / new entries without --update, 2 = script error).

### INFO-2: `--update` runs no longer print the "Run --update" interstitial
`scripts/dep-audit/checksum.py:444-462` — the `**New deps (not yet in DB):**` listing and its "Run `scripts/dep-audit/checksum.py --update` locally" follow-on are now gated on `not args.update`. By the time the report renders under `--update` mode, the entries have already been written; the framing was stale. The trailing `DB updated: +N new` line on stderr remains the authoritative signal of what was added.

## Tests

New file: `scripts/dep-audit/test_checksum.py` (stdlib `unittest` only — **no new deps**, no pytest). Twelve tests:

- `ClassifySourceTests` (5): synthetic `path+file:///tmp/some-path-dep` returns `("path", "/tmp/some-path-dep")`; nested file:// path; registry, git, and workspace (None) sources still classify correctly.
- `PathDepExclusionTests` (2): synthetic tree containing `target/`, `.git/`, `*.swp`, `*.swo`, `*~`, `.DS_Store` — manifest must include `Cargo.toml` and `src/lib.rs` and must NOT include any of the excluded paths. Stronger property test: adding more `target/` and `.git/` noise to the same tree must NOT change the resulting checksum.
- `ExitCodeTwoTests` (3): `classify_source("svn+...")` raises `SystemExit(2)`; `_die(...)` raises `SystemExit(2)`; spawning the script with `--lock /nonexistent` via `subprocess` exits with code 2 and prints `error:` to stderr.
- `UpdateModeInterstitialTests` (2): synthetic minimal `Cargo.lock` with one registry dep; under `--update` the captured stdout must NOT contain "Run \`scripts/dep-audit/checksum.py --update\`", "CI never runs with --update", or "**New deps (not yet in DB):**", but must still contain "STATUS: PASS"; without `--update` the same lock produces the interstitial and exits 1 (regression guard so the gate isn't overzealous).

Run with either:

```
python3 -m unittest discover -s scripts/dep-audit -p test_checksum.py -v
python3 scripts/dep-audit/test_checksum.py
```

## Verification

- 12/12 unit tests pass
- `python3 scripts/dep-audit/checksum.py --help` parses cleanly
- `python3 scripts/dep-audit/checksum.py` against the real repo: 972 deps tracked, 972 matches, STATUS: PASS, exit 0. No drift; all existing checksums still validate (the fixes are byte-compatible for the existing registry/git deps)

## No new dependencies

This PR adds **zero** dependencies. The only new imports in `checksum.py` are `fnmatch` and `urllib.parse`, both Python stdlib. The test file imports `contextlib`, `importlib.util`, `io`, `os`, `subprocess`, `sys`, `tempfile`, `unittest`, `pathlib` — all stdlib. No `pytest`, no third-party packages of any kind.

## CI

The dep-audit workflow will run on push and exercise `checksum.py` end-to-end against the real `Cargo.lock` and committed DB.

### DIFF
--- a/scripts/dep-audit/checksum.py
+++ b/scripts/dep-audit/checksum.py
@@ -51,16 +51,26 @@ scripts/dep-audit/README.md for the operator's guide.
 """
 
 import argparse
+import fnmatch
 import hashlib
 import os
 import re
 import sys
 from pathlib import Path
+from urllib.parse import urlparse
 
 SCRIPT_DIR = Path(__file__).resolve().parent
 REPO_ROOT = SCRIPT_DIR.parent.parent
 DEFAULT_LOCK = REPO_ROOT / "Cargo.lock"
 DEFAULT_DB = REPO_ROOT / ".cryptid" / "dep-audit" / "checksums.db"
+
+# Path-dep manifest exclusions. Anything matching these is omitted from the
+# deterministic file walk so the path-dep checksum doesn't churn on every
+# build. `target/` and `.git/` are directory names (matched against any
+# directory in the walk); the trailing entries are filename glob patterns
+# matched via fnmatch (editor swap files, OS metadata).
+PATH_DEP_EXCLUDE_DIRS = {"target", ".git"}
+PATH_DEP_EXCLUDE_FILE_GLOBS = ("*.swp", "*.swo", "*~", ".DS_Store")
 
 DB_HEADER = (
     "# btt dependency checksum database.\n"
@@ -81,6 +91,21 @@ DB_HEADER = (
 
 
 # ---------------------------------------------------------------------------
+# Error exit helper
+# ---------------------------------------------------------------------------
+
+def _die(msg):
+    """
+    Emit a script-error message to stderr and exit with code 2, per the
+    contract in the module docstring (1 = mismatch/new entry without --update,
+    2 = script error). Raises SystemExit(2) so callers that want to intercept
+    can still do so with `except SystemExit`.
+    """
+    print(msg, file=sys.stderr)
+    raise SystemExit(2)
+
+
+# ---------------------------------------------------------------------------
 # Cargo.lock parser (stdlib-only; Python 3.11 has tomllib but we avoid it
 # to keep this script runnable on any Python 3.x without version checks).
 # ---------------------------------------------------------------------------
@@ -94,7 +119,7 @@ def parse_cargo_lock(path):
     try:
         text = path.read_text(encoding="utf-8")
     except OSError as exc:
-        raise SystemExit(f"error: cannot read {path}: {exc}")
+        _die(f"error: cannot read {path}: {exc}")
 
     packages = []
     current = None
@@ -136,6 +161,11 @@ def classify_source(source):
     """
     Return (kind, rest) where kind is 'registry', 'git', 'path', or 'workspace'
     (the last for packages with no `source` line — the workspace crate itself).
+
+    For path+ deps, Cargo writes `path+file:///abs/path` in Cargo.lock. We
+    parse the file:// URL and return the absolute filesystem path, so the
+    caller receives a value ready to hand to pathlib without re-joining
+    against REPO_ROOT (which would hard-fail on an absolute URL-shaped string).
     """
     if source is None:
         return ("workspace", "")
@@ -144,8 +174,12 @@ def classify_source(source):
     if source.startswith("git+"):
         return ("git", source[len("git+"):])
     if source.startswith("path+"):
-        return ("path", source[len("path+"):])
-    raise SystemExit(f"error: unknown source kind: {source!r}")
+        rest = source[len("path+"):]
+        if rest.startswith("file://"):
+            parsed = urlparse(rest)
+            rest = parsed.path
+        return ("path", rest)
+    _die(f"error: unknown source kind: {source!r}")
 
 
 def checksum_registry(pkg):
@@ -158,7 +192,7 @@ def checksum_registry(pkg):
     tripwire fires.
     """
     if not pkg.get("checksum"):
-        raise SystemExit(
+        _die(
             f"error: registry dep {pkg['name']}@{pkg['version']} has no "
             f"checksum field in Cargo.lock; cannot verify"
         )
@@ -173,13 +207,13 @@ def checksum_git(pkg, rest):
     content hash. We don't need to check out the tree.
     """
     if "#" not in rest:
-        raise SystemExit(
+        _die(
             f"error: git dep {pkg['name']}@{pkg['version']} has no commit "
             f"fragment in source {rest!r}"
         )
     commit = rest.rsplit("#", 1)[1]
     if not re.fullmatch(r"[0-9a-f]{40}", commit):
-        raise SystemExit(
+        _die(
             f"error: git dep {pkg['name']}@{pkg['version']} has non-sha1 "
             f"commit fragment {commit!r}"
         )
@@ -197,21 +231,25 @@ def checksum_path(pkg, rest):
     if not root.is_absolute():
         root = (REPO_ROOT / rest).resolve()
     if not root.exists():
-        raise SystemExit(
+        _die(
             f"error: path dep {pkg['name']}@{pkg['version']} points at "
             f"{root} which does not exist"
         )
     entries = []
     for dirpath, dirnames, filenames in os.walk(root):
+        # Prune excluded directories in place so os.walk does not descend.
+        dirnames[:] = [d for d in dirnames if d not in PATH_DEP_EXCLUDE_DIRS]
         dirnames.sort()
         for fn in sorted(filenames):
+            if any(fnmatch.fnmatch(fn, pat) for pat in PATH_DEP_EXCLUDE_FILE_GLOBS):
+                continue
             full = Path(dirpath) / fn
             if full.is_symlink():
                 kind = "l"
                 try:
                     target = os.readlink(full)
                 except OSError as exc:
-                    raise SystemExit(f"error: cannot read symlink {full}: {exc}")
+                    _die(f"error: cannot read symlink {full}: {exc}")
                 content_hash = hashlib.sha256(target.encode("utf-8")).hexdigest()
             else:
                 kind = "f"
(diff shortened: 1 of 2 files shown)


moorkh (commented):
[cryptid] barbarian review, PR #47 (`cryptid/issue-22-checksum-followups`, a92331c). Python-only, no cargo invoked.

## Verdict: APPROVE

All four follow-ups land correctly. Tests pass 12/12. Real-repo validate holds at 972/972 matches, PASS, rc=0. `--update` runs clean and suppresses the stale interstitial as designed. Mutation probes confirm each of the three load-bearing assertions actually fail when the protection is removed.

## What I ran

- `python3 -m unittest discover -s scripts/dep-audit -p test_checksum.py -v` -> 12 tests, OK
- `python3 scripts/dep-audit/checksum.py --help` -> parses
- `python3 scripts/dep-audit/checksum.py` -> 972 tracked, 972 matches, 0 new, 0 stale, 0 mismatches, STATUS: PASS, rc=0
- `python3 scripts/dep-audit/checksum.py --update` -> same, no interstitial, no DB write (nothing new)

## Findings by item

### LOW-1 path+file:// URL handling

`classify_source` only routes to `urlparse` when `rest.startswith("file://")`, and uses `parsed.path`. I probed urlparse directly with `path+file:///tmp/foo` and confirmed it returns `scheme='path+file', netloc='', path='/tmp/foo'` — so the code path of parsing `rest` (= `file:///tmp/foo`) yields `path='/tmp/foo'` correctly. `test_path_file_url_with_nested_path` covers a real nested path `/home/dev/projects/my-crate`.

Three edge cases I'd note but not block on (all out of scope for Linux CI today):

1. `path+file://localhost/tmp/foo` — urlparse sets `netloc='localhost'` and `path='/tmp/foo'`. The code silently drops `netloc`. Result is still a usable `/tmp/foo`, so it works for the one tool that emits the localhost form, but we don't assert anything about it.
2. `path+file:///tmp/foo?x=1` or `#frag` — query/fragment silently discarded via `parsed.path`. A path dep should never carry a query; a defensive reject (via `_die`) would be slightly stronger. Not blocking.
3. Windows `path+file:///C:/foo` -> urlparse path `/C:/foo` has a stray leading slash. CI is Linux; file as future-proofing only.

None of these are regressions; the old code was strictly worse. Accept as-is.

### LOW-2 path-dep exclude list

The walker correctly uses the `dirnames[:] = [...]` in-place prune idiom (checksum.py:241), so `os.walk` does not descend into excluded directories — wasted I/O avoided, not just post-filtered. Filename filtering uses `fnmatch.fnmatch` in a generator expression (line 244); `fnmatch.filter` would be marginally more efficient but the difference is noise against the file read cost. Accept.

`PATH_DEP_EXCLUDE_DIRS = {"target", ".git"}` is matched against the directory basename in `dirnames`, so a nested `subcrate/target/` inside a workspace path dep is also excluded. Correct.

`test_manifest_stable_when_only_excluded_files_change` does exactly what the PR description claims: digest before, mutate target/.git/swap noise, digest after, assert equal. Mutation probe: I set `PATH_DEP_EXCLUDE_DIRS = set()` and `PATH_DEP_EXCLUDE_FILE_GLOBS = ()` then re-ran the tree — `target/debug/build.log` appeared in the manifest and the two digests diverged. The test is load-bearing.

One curiosity I probed but will not block on: a file literally named `target` (not a directory) would NOT be excluded, because `os.walk` only offers basename-prune for dirs; files are matched only against the glob list. This is the correct behaviour — `target/` the build dir should go, `target` the file should stay. No action.

### INFO-1 exit code 2 on script error

Grepped `checksum.py` for `raise SystemExit\(` post-PR: exactly one occurrence, inside `_die` itself (line 105). All 9 historical error sites route through `_die`. `_die` prints to `sys.stderr` (line 104) and raises `SystemExit(2)` (line 105). Call sites still pass `f"error: ..."` strings, so the on-wire format is unchanged — anything grepping for `error:` in CI logs keeps working.

`test_unreadable_lock_exits_two_subprocess` launches a real subprocess via `subprocess.run([sys.executable, ...])` — not a mock. That is the right call; it catches the `_die` path through actual argparse wiring and module loading. `test_die_helper_exits_two` additionally pins the helper itself. `test_unknown_source_kind_exits_two_in_process` pins `classify_source`'s unknown-kind branch.

Mutation probe: I monkeypatched `_die` to `raise SystemExit(1)`. `test_die_helper_exits_two` and `test_unknown_source_kind_exits_two_in_process` both fail as expected (exit 1 != 2). Load-bearing.

Note: `main()` at line 371 still has a non-`_die` error path (a bare `print(...) return 2`) for the general `parse_cargo_lock` Exception wrapper. That is fine — it's inside `main()` where `return 2` is the idiomatic form and cannot use `_die` because the intent is to continue unwinding to the `sys.exit(main())` at the bottom, not raise from deep inside a try block. Consistent exit code.

### INFO-2 --update interstitial

Gate at line 449 is `if new_entries and not args.update:` — straightforward. The trailing `DB updated: +N new, -N stale` line at lines 506-510 is unconditional inside the `if args.update:` branch, so it still fires.

`test_update_run_omits_run_update_interstitial` asserts three things: no `Run \`scripts/dep-audit/checksum.py --update\`` string, no `CI never runs with --update`, no `**New deps (not yet in DB):**` header — and also asserts the DB was written. `test_non_update_run_keeps_run_update_interstitial` is the positive control.

Mutation probe: ran `main(['--lock', ...])` (non-update) against a synthetic lock — the interstitial DID appear. In other words, if the gate were removed the `--update` run would match the non-update output and the test would trip on the `assertNotIn('**New deps (not yet in DB):**')` line. Load-bearing.

One minor gap: there's no explicit test for `--update` with zero new deps (the "empty gated block" path). I exercised it manually against the real repo: `python3 scripts/dep-audit/checksum.py --update` printed `STATUS: PASS` with no interstitial and no `DB updated:` line (correct, since nothing was new or stale). Adding a unit test for this would be belt-and-braces; not blocking.

## Scope and hygiene

- `git diff --stat HEAD~1 HEAD`: exactly `scripts/dep-audit/checksum.py` (+82/-20) and `scripts/dep-audit/test_checksum.py` (+322). No scope creep.
- New imports in `checksum.py`: `fnmatch` (stdlib), `from urllib.parse import urlparse` (stdlib). No third-party deps. Within dependency discipline.
- `test_checksum.py` imports: `contextlib`, `importlib.util`, `io`, `os`, `subprocess`, `sys`, `tempfile`, `unittest`, `pathlib.Path`, `fnmatch`. All stdlib.
- No `__pycache__` or `.pyc` in the commit. (`scripts/dep-audit/__pycache__/` is present locally from my test run but untracked, as expected.)
- Tests pass in isolation and together (single `unittest discover` invocation covers both).

## Verdict

APPROVE. Ship it. Do not merge (self-PR; Sergei or a human to merge).

Three optional follow-ups I would NOT block on, left here as breadcrumbs in case anyone wants to tighten further:

1. `classify_source` could reject `path+file://` URLs with non-empty `query`/`fragment` via `_die`, rather than silently dropping them. Defensive.
2. Add a unit test for `--update` with zero new deps and zero stale — pins the "empty gated block + no DB write" path explicitly.
3. Consider warning on non-empty `netloc` in `path+file://` (the localhost form) — lossless today, would help if cargo ever changed its emission.

None of the above are in scope for #22.

PR #48, opened 2026-04-14
[cryptid] paths.rs: batched follow-ups from PR #36 barbarian (#37)

### LINKED ISSUE #37: [cryptid] paths.rs follow-ups from PR #36 barbarian
## Context

Five LOW / cosmetic findings from the PR #36 barbarian review (merge commit efda847). All non-blocking, all worth tracking so they don't get lost.

## 1. `seat_home_env` env-restoration asymmetry

In `src/commands/wallet_keys.rs` tests, `seat_home_env` sets both `HOME` AND `XDG_CONFIG_HOME` (or the per-OS equivalent), but the test save/restore helpers only persist `HOME` across the test boundary. Result: `XDG_CONFIG_HOME` leaks into the rest of the process between tests.

**Why it doesn't bite today:** every env-touching test acquires `HOME_LOCK` / `ENV_LOCK` and unconditionally re-sets the vars before running, so the leaked state gets clobbered before any assertions. But the moment a test forgets to re-set `XDG_CONFIG_HOME` (or a new test is added that depends on it being unset), the leak becomes a flaky-test source.

**Fix:** make `seat_home_env` return a guard that restores BOTH vars on drop, not just `HOME`. Symmetric with the existing `HOME` save/restore.

## 2. `legacy_fallback_when_only_old_dir_exists` doesn't assert stderr emission

The test confirms `config_dir()` returns the legacy path when only `$HOME/.bittensor` exists, but does NOT assert that the once-per-process stderr warning was actually emitted. If a future refactor accidentally drops the `warn_legacy` call, the test still passes silently.

**Fix:** capture stderr (via a redirected `Write` or by spawning a subprocess) and assert the warning text appears.

## 3. No test for the `Once` semantics

The `std::sync::Once` guard means the legacy warning fires at most once per process. There's no test that verifies this — calling `config_dir()` twice with the legacy path in place should produce exactly one stderr line, not two. Without a test, a future refactor that replaces `Once` with a free-standing call would silently double-warn.

**Fix:** test that asserts the warning fires exactly once across N consecutive `config_dir()` calls. Probably needs a subprocess to get a fresh `Once` per test.

## 4. `cfg(target_os)` logic duplicated across three sites

`paths::native_config_dir`, `test_wallets_parent` (in wallet_keys tests), and `seat_home_env` (also in tests) all branch on `cfg(target_os)`. Three sites means three places to update if the resolver rules ever change (e.g., adopting macOS's `Library/Caches` for ephemeral data, or honoring `%LOCALAPPDATA%` on windows).

**Fix:** consolidate into a single per-OS resolver in `paths.rs` exposed via a small test-helper API. The test helpers should call into the same code path the production resolver uses, just with overridable inputs.

## 5. `VarError::NotUnicode` collapsed with `NotPresent`

`config_dir()` returns the same "HOME not set" / "APPDATA not set" error for both `VarError::NotPresent` and `VarError::NotUnicode`. The latter case is rare on linux (almost never on macOS / windows) but the diagnostic could be more helpful: "HOME contains non-UTF-8 bytes" tells the user to fix their shell config, not to set the var.

**Fix:** match on `VarError` explicitly and return distinct error messages.

## Severity

All LOW. None of these affect correctness on any platform that's actually targeted (linux is the primary, macOS is nice-to-have, windows is out of scope until someone needs it). The barbarian APPROVED PR #36 without blocking on any of them.

## Acceptance

- `seat_home_env` returns a symmetric guard restoring both vars
- `legacy_fallback` test asserts on stderr emission
- New `Once`-semantics test exists
- `cfg(target_os)` consolidated to one site
- `VarError` arms produce distinct messages

Can be addressed as a single PR or split into smaller ones, builder's call.

moorkh:
Closes #37.

Five LOW / cosmetic items from the PR #36 barbarian review, batched into one PR per builder's discretion. All non-blocking, no behavior change on the happy path. No new dependencies.

### 1. Symmetric env guard (`paths::EnvGuard`)
`src/commands/paths.rs:243-319`, `src/commands/wallet_keys.rs:1633-1671`

`EnvGuard::capture()` snapshots `HOME`, `XDG_CONFIG_HOME`, and `APPDATA` together, and `Drop` restores all three. `seat_config_env_at(tmp)` returns the guard alongside the computed `wallets/` parent. The old `seat_home_env` (wallet_keys) and `seat_home`/`restore_home` (wallet_keys) helpers now delegate to it, replacing the asymmetric `original_home` save/restore that leaked `XDG_CONFIG_HOME` between tests.

### 2. `warn_legacy` writer injection + banner assertion
`src/commands/paths.rs:148-179`

`warn_legacy_to(&mut dyn Write, &Once, legacy, new)` is the core implementation. `warn_legacy` is a thin wrapper that pins `stderr().lock()` and the process-wide `LEGACY_WARNING` static. This required refactoring `warn_legacy` to take a writer — option 3 from the issue brief (no new dep, no subprocess spawn). `warn_legacy_writes_banner_to_writer` now asserts the exact banner text end-to-end.

### 3. `Once`-semantics test
`src/commands/paths.rs:148-179` (refactor), test at `warn_legacy_fires_exactly_once_per_once`

Because `warn_legacy_to` takes `&Once`, the test supplies a fresh one, hits it twice, and asserts the second call writes zero bytes. A third call with a *different* fresh `Once` does emit — so future refactors that drop the guard or replace `Once` with an unconditional call will fail the test.

### 4. Single per-OS resolver site
`src/commands/paths.rs:56-121` (`config_dir_from_env`), `src/commands/paths.rs:236-274` (test helpers), `src/commands/wallet_keys.rs:1673-1678`

`config_dir_from_env(home, xdg, appdata) -> Result<PathBuf, BttError>` is the single source of truth for `cfg(target_os)` branching (mirrors the `expand_tilde` refactor shape from PR #45). `native_config_dir` reads env vars and dispatches. `wallets_parent_for` (pub(crate) test helper) also dispatches through it. The old `test_wallets_parent` duplicate in `wallet_keys::tests` is now a thin delegation. Three cfg sites → one.

### 5. `VarError::NotUnicode` distinct messages
`src/commands/paths.rs:37-49` (`read_env_var`), distinct tests at `linux_home_not_present_has_distinct_message` and `linux_home_not_unicode_has_distinct_message`

`read_env_var` matches on `VarError` explicitly: `NotPresent` → `"X environment variable not set"`, `NotUnicode(_)` → `"X contains non-UTF-8 bytes"`. Used by `native_config_dir` and `legacy_dir`.

## Tests added

Nine new tests in `paths::tests`:

- `env_guard_restores_all_three_vars_on_drop` — set sentinels, seat, drop, assert all three restored
- `env_guard_restores_unset_vars_to_unset` — unset → seat → drop → assert all three unset
- `warn_legacy_writes_banner_to_writer` — asserts full banner text into a `Vec<u8>`
- `warn_legacy_fires_exactly_once_per_once` — `Once` semantics, plus fresh-`Once` re-emission check
- `linux_home_not_present_has_distinct_message` — exercises `NotPresent` arm
- `linux_home_not_unicode_has_distinct_message` — sets a `0xFF`-containing `OsString` for HOME, exercises `NotUnicode` arm
- `resolver_linux_xdg_wins` — synthetic values, no env mutation
- `resolver_linux_empty_xdg_falls_back_to_home` — synthetic
- `resolver_linux_no_xdg_uses_home` — synthetic
- `resolver_linux_missing_home_errors` — synthetic
- `resolver_macos_uses_application_support` (cfg-gated)
- `resolver_windows_uses_appdata` (cfg-gated)

## Verification

- `cargo build` — ok
- `cargo clippy --all-targets --all-features -- -D warnings` — clean
- `cargo test --release paths` — 15 passed, 0 failed
- `cargo test --release wallet_keys` — 69 passed, 1 ignored, 0 failed

## No new deps

Confirmed. `Cargo.toml` unchanged.

### DIFF
--- a/src/commands/paths.rs
+++ b/src/commands/paths.rs
@@ -22,85 +22,167 @@
 //! warning telling them how to migrate. btt never silently moves wallet
 //! material on the user's behalf.
 
-use std::path::PathBuf;
+use std::env::VarError;
+use std::io::Write;
+use std::path::{Path, PathBuf};
 use std::sync::Once;
 
 use crate::error::BttError;
 
-/// Legacy wallet-root location used by btt prior to this module.
+/// Read an environment variable, distinguishing "not set" from "not UTF-8".
 ///
-/// This is the btcli-compatible path. It is only consulted as a fallback
-/// when the new OS-dependent config directory does not yet exist.
-fn legacy_dir() -> Option<PathBuf> {
-    let home = std::env::var("HOME").ok()?;
-    if home.is_empty() {
-        return None;
+/// Returns:
+/// - `Ok(Some(value))` if the variable is set and valid UTF-8.
+/// - `Ok(None)` if the variable is not present.
+/// - `Err(BttError)` if the variable is present but contains non-UTF-8 bytes.
+fn read_env_var(key: &str) -> Result<Option<String>, BttError> {
+    match std::env::var(key) {
+        Ok(v) => Ok(Some(v)),
+        Err(VarError::NotPresent) => Ok(None),
+        Err(VarError::NotUnicode(_)) => Err(BttError::io(format!(
+            "{} contains non-UTF-8 bytes",
+            key
+        ))),
     }
-    Some(PathBuf::from(home).join(".bittensor"))
 }
 
-/// Compute the OS-native config directory without consulting the filesystem.
-fn native_config_dir() -> Result<PathBuf, BttError> {
+/// Compute the OS-native config directory from explicit env values.
+///
+/// This is the single source of truth for the per-OS resolver logic.
+/// Production callers go through [`config_dir`] / [`native_config_dir`],
+/// which read the process environment; tests (and the wallet_keys test
+/// helpers) can call this directly with synthetic values to avoid any
+/// `cfg(target_os)` branching of their own.
+///
+/// Inputs:
+/// - `home`: value of `$HOME`. Required on linux / macOS / BSDs.
+/// - `xdg`: value of `$XDG_CONFIG_HOME`. Consulted only on linux / BSDs.
+/// - `appdata`: value of `%APPDATA%`. Required on windows.
+///
+/// An `Some("")` empty-string input is treated as "not set" (matching the
+/// historical behavior of `native_config_dir`), so callers can pass raw
+/// env-var reads straight through.
+pub(crate) fn config_dir_from_env(
+    home: Option<&str>,
+    xdg: Option<&str>,
+    appdata: Option<&str>,
+) -> Result<PathBuf, BttError> {
     #[cfg(target_os = "linux")]
     {
-        if let Ok(xdg) = std::env::var("XDG_CONFIG_HOME") {
-            if !xdg.is_empty() {
-                return Ok(PathBuf::from(xdg).join("btt"));
+        let _ = appdata;
+        if let Some(x) = xdg {
+            if !x.is_empty() {
+                return Ok(PathBuf::from(x).join("btt"));
             }
         }
-        let home = std::env::var("HOME")
-            .map_err(|_| BttError::io("HOME environment variable not set"))?;
-        Ok(PathBuf::from(home).join(".config").join("btt"))
+        let h = home
+            .filter(|s| !s.is_empty())
+            .ok_or_else(|| BttError::io("HOME environment variable not set"))?;
+        Ok(PathBuf::from(h).join(".config").join("btt"))
     }
     #[cfg(target_os = "macos")]
     {
-        let home = std::env::var("HOME")
-            .map_err(|_| BttError::io("HOME environment variable not set"))?;
-        Ok(PathBuf::from(home)
+        let _ = xdg;
+        let _ = appdata;
+        let h = home
+            .filter(|s| !s.is_empty())
+            .ok_or_else(|| BttError::io("HOME environment variable not set"))?;
+        Ok(PathBuf::from(h)
             .join("Library")
             .join("Application Support")
             .join("btt"))
     }
     #[cfg(target_os = "windows")]
     {
-        let appdata = std::env::var("APPDATA")
-            .map_err(|_| BttError::io("APPDATA environment variable not set"))?;
-        Ok(PathBuf::from(appdata).join("btt"))
+        let _ = home;
+        let _ = xdg;
+        let a = appdata
+            .filter(|s| !s.is_empty())
+            .ok_or_else(|| BttError::io("APPDATA environment variable not set"))?;
+        Ok(PathBuf::from(a).join("btt"))
     }
     #[cfg(not(any(target_os = "linux", target_os = "macos", target_os = "windows")))]
     {
+        let _ = appdata;
         // BSDs, Solaris, illumos, etc.: default to the XDG-style linux
         // layout. Users on exotic targets can set `XDG_CONFIG_HOME`
         // explicitly if they want a different location.
-        if let Ok(xdg) = std::env::var("XDG_CONFIG_HOME") {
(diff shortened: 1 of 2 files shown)


moorkh (commented):
[cryptid] barbarian review — APPROVE

Ran the worktree locally on commit 136ade1:
- cargo build: pass
- cargo clippy --all-targets --all-features -- -D warnings: clean
- cargo test --release paths: 15 passed, 0 failed
- cargo test --release wallet_keys: 69 passed, 0 failed, 1 ignored
- git diff --stat HEAD~1 HEAD: only paths.rs (+510/-49) and wallet_keys.rs (+36/-93). Cargo.toml / Cargo.lock untouched.

Mutation probes (all three killed the relevant tests):

1. EnvGuard::drop patched to skip XDG restore — env_guard_restores_all_three_vars_on_drop fails with `left: "/tmp/.../xdg", right: "/orig/xdg"`. Symmetric restore is actually verified.
2. warn_legacy_to patched to ignore `&mut dyn Write` and write via eprintln! — warn_legacy_writes_banner_to_writer fails on `assertion failed: s.contains("legacy wallet directory at ...")`. Writer injection is actually exercised.
3. warn_legacy_to patched to ignore the `&Once` parameter — warn_legacy_fires_exactly_once_per_once fails on `second call on same Once must emit nothing`. Gate is actually checked inside the function.

cfg(target_os) sites in production: exactly one function (config_dir_from_env) holds all four dispatch branches. native_config_dir and wallets_parent_for are cfg-free. Consolidation is complete.

Specific checks:
- Empty-XDG edge case preserved: `if !x.is_empty()` at line 74; resolver_linux_empty_xdg_falls_back_to_home covers it.
- Zero `.unwrap()` on production paths.
- warn_legacy_to writes to `Vec<u8>` which is infallible — no flush needed; banner matches previous byte layout except for one re-indented writeln! arg (cosmetic, not semantic).
- Once-semantics test asserts `Vec::is_empty()` (buffer-len check), not substring absence. Correct.
- NotUnicode test properly gated `#[cfg(target_os = "linux")]` and uses `OsStringExt::from_vec([..., 0xFF])`. Passes on linux, no-ops elsewhere.
- No new BttError variants; no callers pattern-matched on the old single message, so no break.

Latent observations (non-blocking, worth a future follow-up):

a. EnvGuard stores `Option<String>` not `Option<OsString>`. If the process has a non-UTF-8 HOME at construction time, `std::env::var("HOME").ok()` returns None and Drop will `remove_var` instead of restoring the original non-UTF-8 value. PR #36's item 1 explicitly called OsString "more correct"; implementation went with String. Harmless in CI (UTF-8 env) but the asymmetry is real. Trivial fix: use `std::env::var_os()`. Not blocking — the test-only surface makes this theoretical.

b. read_env_var returns `BttError::io(...)` for both NotPresent-at-callsite and NotUnicode. The messages are distinct ("HOME environment variable not set" vs "HOME contains non-UTF-8 bytes"), but they are the same variant. Callers cannot programmatically branch. The PR description wording "distinct error variants" is stronger than what the code delivers; "distinct error messages" is accurate. Sufficient for a LOW follow-up; flagging so a future consumer who expects a typed distinction isn't surprised.

c. config_dir_from_env grew a `#[cfg(not(any(linux, macos, windows)))]` fallback block that gives BSDs the XDG-style resolver. That's a new behavior not mentioned in the PR body — pre-PR, BSD targets presumably didn't compile (no cfg arm) or hit some other path. Worth a one-liner in the PR description for the historian. Not a defect.

Verdict: APPROVE. The follow-ups are correctly implemented, tests are not vacuous (mutation probes confirmed), and the consolidation work is complete. Observations a–c are sub-LOW and should not block merge; they can land as a future ticket if anyone cares.

PR #49, opened 2026-04-14
[cryptid] wallet cleanup: reap stale .tmp.*, .bak.*, .lock.* (#42)

### LINKED ISSUE #42: [cryptid] wallet cleanup: reap stale .tmp.* and .bak.* directories

moorkh:
Closes #42.

## Summary

Adds `btt wallet cleanup [--dry-run] [--wallet <name>] [--older-than <duration>]` to reap the three reserved-prefix artefacts left under `<wallets>/` by crashed or interrupted `wallet create` runs:

- `.tmp.<name>.<pid>.<nanos>.<ctr>/` staging dirs (PR #40)
- `.bak.<name>.<pid>.<nanos>.<ctr>/` backup dirs (PR #40)
- `.lock.<name>` per-wallet `flock(2)` sentinels (PR #43)

## Grammar (hand-rolled parser, no `regex` dep)

`src/commands/wallet.rs::parse_reserved_entry` matches exactly:

    .tmp.<name>.<pid>.<nanos>.<ctr>
    .bak.<name>.<pid>.<nanos>.<ctr>
    .lock.<name>

where

- `<name>` is 1..=64 chars of `[A-Za-z0-9_-]` (ASCII only, no `.`, no `/`, no NUL).
- `<pid>` is `\d+` (non-empty ASCII digit run).
- `<nanos>` is `\d{1,19}` (u64 nanos-since-epoch, 19 digits max).
- `<ctr>` is `\d+`.

The parser splits `<name>.<pid>.<nanos>.<ctr>` from the right by dots — since `<name>` cannot contain `.` (enforced in `is_valid_wallet_name`), the last three segments are the numeric tail and everything before them is `<name>`. Any entry that starts with a reserved prefix but fails this grammar is reported as `skipped-no-match` and never touched.

## Lock-probe safety check (moorkh's comment on #42)

`.lock.*` files require a non-blocking `flock(LOCK_EX | LOCK_NB)` probe before unlink. Logic lives in a new `cleanup_flock` module local to `wallet.rs`:

- Open the lock file `O_RDWR`.
- `flock(fd, LOCK_EX | LOCK_NB)`.
- If `rc == 0`: lock was free. Release via `LOCK_UN` immediately, then `unlink(2)`. Action = `reaped`.
- If `errno == EWOULDBLOCK/EAGAIN` (linux: 11; macOS: 35): lock is held by another process. Action = `skipped-held`. **Never unlinked.**
- Any other errno: propagate as a hard error — silently treating a failed probe as "unheld" would defeat the check.

This closes the race moorkh flagged: cleanup must not remove a lock file that a concurrent `wallet create` is actively using to serialize its backup-and-rename sequence.

## Where `LOCK_NB` / `flock_sys` landed

**New local module `cleanup_flock` in `wallet.rs`**, not a shared helper. The two paths want different shapes:

- `wallet_keys::flock_sys::LockGuard::acquire` takes ownership of a `File` and holds a blocking `LOCK_EX` until drop — it guards the create critical section.
- `cleanup_flock::probe_unheld` is a one-shot non-blocking probe that answers "is anyone holding this lock right now?" and releases the lock before returning so the subsequent unlink does not race a would-be acquirer.

Merging them into a single guard type would require a mode parameter threaded through the create critical section and would make the wallet_keys hot path harder to reason about. The duplication cost is a three-line `extern \"C\"` block plus three integer constants (`LOCK_EX = 2`, `LOCK_UN = 8`, `LOCK_NB = 4`).

## `--older-than` parser

Hand-rolled `parse_older_than` accepts `\d+[smhd]` (e.g. `60s`, `30m`, `24h`, `7d`). No `humantime` dep. Overflow-checked via `checked_mul`. Rejects empty input, missing unit, non-digit leading part, fractional values, and negative signs.

## Safety invariants enforced

1. **Pattern match exact.** Only entries that parse per the grammar are candidates. Lookalikes like `.tmp.fooblahblah` are `skipped-no-match`, never reaped.
2. **Lock probe.** `.lock.*` files are flock-probed before unlink.
3. **Symlinks not followed.** `fs::symlink_metadata` + `file_type().is_symlink()` check before any `remove_dir_all` / `remove_file`. A symlink matching the pattern is `skipped-no-match` even if its target is a real dir.
4. **Top-level only.** Scan is the immediate children of `<wallets>/`. No recursion.
5. **`--wallet <name>` validated.** Rejected if not 1..=64 chars of `[A-Za-z0-9_-]`. Path-traversal (`..`, `/`), NUL bytes, and anything outside the grammar are refused with a clear error before the scan runs.

## JSON output schema

    {
      \"ok\": true,
      \"data\": {
        \"entries\": [
          {
            \"path\": \"/home/alice/.config/btt/wallets/.tmp.foo.1234.1700000000000.0\",
            \"kind\": \"tmp\",
            \"action\": \"reaped\"
          },
          {
            \"path\": \"/home/alice/.config/btt/wallets/.lock.bar\",
            \"kind\": \"lock\",
            \"action\": \"skipped-held\"
          },
          {
            \"path\": \"/home/alice/.config/btt/wallets/.bak.baz.99.1.0\",
            \"kind\": \"bak\",
            \"action\": \"skipped-too-young\"
          }
        ]
      }
    }

`kind` is one of `tmp` / `bak` / `lock`. `action` is one of `reaped` / `kept-dry-run` / `skipped-held` / `skipped-too-young` / `skipped-no-match`. Entries are sorted by path for deterministic consumption by `jq`.

## Tests added

11 end-to-end cleanup scenarios in `src/commands/wallet.rs::tests` (all `HOME_LOCK`-guarded; unix-only where required):

1. `cleanup_reaps_stale_tmp_dir` — plant `.tmp.foo.123.456789.0/` with inner files, assert it's gone and reported as `reaped`.
2. `cleanup_reaps_stale_bak_dir` — same shape for `.bak.*`.
3. `cleanup_reaps_unheld_lock_file` — plant `.lock.foo`, assert it's gone and reported as `reaped`.
4. `cleanup_skips_held_lock_file` — open `.lock.foo` and acquire `LOCK_EX` directly via a local `extern \"C\" flock` (no dependency on `wallet_keys::flock_sys` internals), run cleanup, assert the file is **not** gone and is reported as `skipped-held`. Releases the lock at end.
5. `cleanup_skips_real_wallet` — plant a minimal real wallet, run cleanup, assert the real wallet is untouched and does **not** appear in the report at all (it doesn't match any reserved prefix).
6. `cleanup_skips_lookalike_dir` — plant `.tmp.fooblahblah` (no numeric tail triple), assert it's **not** removed and reported as `skipped-no-match`.
7. `cleanup_dry_run_reaps_nothing` — plant a `.tmp.*`, run `--dry-run`, assert the dir still exists and is reported as `kept-dry-run`.
8. `cleanup_wallet_filter` — plant `.tmp.foo.*` and `.tmp.bar.*`, run with `--wallet foo`, assert only the `foo` one is reaped.
9. `cleanup_older_than_filter` — plant a fresh `.tmp.*` and an `.bak.*` with mtime backdated 1 hour (via `File::set_modified`), run with `--older-than 30m`, assert the bak is `reaped` and the tmp is `skipped-too-young`.
10. `cleanup_rejects_traversal_in_wallet_name` — invoke with `--wallet ../etc` (and separately with `foo\\0bar`), assert both error with a wallet-name-validation message.
11. `cleanup_skips_symlink` — plant `.tmp.foo.1.2.3` as a symlink pointing **outside** the wallets dir at a sentinel with a canary file, run cleanup, assert the symlink is **not** removed, the canary file at the target is still present (i.e. the symlink was not followed), and the report shows `skipped-no-match`.

Plus 9 parser / duration / name-validation unit tests that don't touch disk or env: canonical tmp/bak/lock parses, lookalike rejection, bad `<pid>`, 20-digit `<nanos>` rejection, unrelated prefix rejection, duration parser units, duration parser rejections, and `is_valid_wallet_name` edge cases.

## Decisions / notes

- **No new deps.** Hand-rolled parser for both the reserved-prefix grammar and `--older-than`. `regex`, `humantime`, `chrono`, `walkdir`, `fs_extra` all deliberately avoided. `Cargo.toml` unchanged.
- `cleanup_flock` module is local to `wallet.rs` (not a shared helper) — see rationale above.
- `--wallet <name>` wallet-name validation is stricter (`[A-Za-z0-9_-]{1,64}`) than the (absent) validation in `wallet_keys::create`, because cleanup is the one path that unlinks things it sees on disk, so ambiguous names are refused on principle.
- The JSON report sorts entries by path for deterministic script consumption.
- No auto-run on startup. Cleanup is always explicit.

## Verification

- `cargo build` — clean.
- `cargo test --release wallet` — 91 passed, 0 failed, 1 ignored (67 filtered). All 11 cleanup end-to-end tests pass alongside the existing wallet / wallet_keys / flock concurrency tests. This single invocation exercises both \`wallet\` and \`wallet_keys\` because the filter is substring-match.

## Test plan

- [ ] CI clippy + build green
- [ ] CI wallet tests green
- [ ] Manual smoke: plant a fake `.tmp.foo.1.2.3` under a throwaway btt config dir and run `btt wallet cleanup --dry-run` / `btt wallet cleanup` and verify the JSON output shape

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,9 @@ btt chain balance <ss58_address>
 # List local wallets
 btt wallet list
 
+# Reap stale staging/backup/lock artefacts from crashed wallet create runs
+btt wallet cleanup [--dry-run] [--wallet <name>] [--older-than 7d]
+
 # Emit SKILL.md for AI agent integration
 btt skill
 ```
@@ -102,6 +105,42 @@ to reconstruct the old wallet from the command's own inputs, so the
 refusal error explicitly warns about irreversibility. If you only want to
 replace one half of a wallet, use `new-coldkey` / `new-hotkey` /
 `regen-coldkey` / `regen-hotkey` instead.
+
+## Wallet cleanup
+
+`wallet create`'s atomic staging path (PR #40) writes sibling
+`.tmp.<name>.<pid>.<nanos>.<ctr>/` and `.bak.<name>.<pid>.<nanos>.<ctr>/`
+directories under `<wallets>/` during a create, and the per-wallet
+`flock(2)` path (PR #43) writes a `.lock.<name>` sentinel file. All three
+prefixes are reserved — `wallet create` refuses them as names, and
+`wallet list` filters them out.
+
+On a crashed or interrupted run, stale `.tmp.*` / `.bak.*` directories
+accumulate. `btt wallet cleanup` is the explicit, opt-in sweep:
+
+```bash
+# List (and reap) stale entries under <wallets>/
+btt wallet cleanup
+
+# Same, but don't remove anything — just emit the JSON report.
+btt wallet cleanup --dry-run
+
+# Reap only the staging/backup/lock entries belonging to a specific wallet.
+btt wallet cleanup --wallet alice
+
+# Reap only entries whose mtime is older than a duration (s/m/h/d).
+btt wallet cleanup --older-than 7d
+```
+
+The command uses a strict reserved-prefix grammar match — it will never
+`remove_dir_all` anything that does not fit `.tmp.<name>.<pid>.<nanos>.<ctr>`
+(or the `.bak.` / `.lock.` analogues). Symlinks are never followed.
+`.lock.*` files are probed with a non-blocking `flock(LOCK_EX | LOCK_NB)`
+before unlink so that a lock currently held by a concurrent `wallet
+create` is reported as `skipped-held` and left on disk. The JSON output
+is `{ok: true, data: {entries: [{path, kind, action}, ...]}}` where
+`kind` ∈ `tmp` / `bak` / `lock` and `action` is one of `reaped`,
+`kept-dry-run`, `skipped-held`, `skipped-too-young`, `skipped-no-match`.
 
 ## btcli format compatibility
 
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -237,6 +237,39 @@ pub enum WalletAction {
         #[arg(long)]
         ss58: String,
     },
+
+    /// Reap stale `.tmp.*`, `.bak.*`, and `.lock.*` entries under the
+    /// wallets directory. These are reserved prefixes left behind by
+    /// `wallet create` on crashed or interrupted runs (see issue #42).
+    ///
+    /// `.tmp.<name>.<pid>.<nanos>.<ctr>/` are atomic-create staging
+    /// dirs, `.bak.<name>.<pid>.<nanos>.<ctr>/` are `--force` backup
+    /// dirs, and `.lock.<name>` are per-wallet `flock(2)` sentinels.
+    /// The cleanup command uses a strict grammar match and will never
+    /// touch anything that does not fit the reserved-prefix pattern.
+    /// `.lock.*` files are probed with a non-blocking `flock(LOCK_NB)`
+    /// and skipped if currently held by another process.
+    ///
+    /// Emits a JSON array of `{path, kind, action}` records.
+    Cleanup {
+        /// List what would be removed without touching disk. Each
+        /// candidate is reported with `action = "kept-dry-run"`.
+        #[arg(long, default_value_t = false)]
+        dry_run: bool,
+        /// Restrict the scan to entries whose `<name>` component
+        /// matches this wallet name verbatim: `.tmp.<name>.*`,
+        /// `.bak.<name>.*`, and `.lock.<name>`. The name is validated
+        /// (1..=64 chars of `[A-Za-z0-9_-]`) before use; path-traversal
+        /// characters (`/`, `..`, NUL) are rejected.
+        #[arg(long, value_name = "NAME")]
+        wallet: Option<String>,
+        /// Only reap entries whose modification time is older than
+        /// this duration. Grammar: `\d+[smhd]` — for example `60s`,
+        /// `30m`, `24h`, `7d`. Default: no age filter (reap everything
+        /// matching the prefix).
+        #[arg(long, value_name = "DURATION")]
+        older_than: Option<String>,
+    },
 }
 
 #[derive(Subcommand, Debug)]
--- a/src/commands/skill.rs
+++ b/src/commands/skill.rs
@@ -89,6 +89,16 @@ btt wallet sign --name <name> --message "<msg>" [--use-hotkey] \
 
 # Verify a signature against an SS58 address
 btt wallet verify --message "<msg>" --signature 0x... --ss58 <address>
+
+# Reap stale `.tmp.*`, `.bak.*`, and `.lock.*` entries left under the
+# wallets directory by crashed or interrupted `wallet create` runs.
+# Uses a strict reserved-prefix grammar match, never follows symlinks,
+# and probes `.lock.*` files with a non-blocking `flock(LOCK_NB)` so a
+# held lock (e.g. from a concurrent `wallet create`) is never unlinked.
+# Emits a JSON array of `{path, kind, action}` entries where `kind` is
+# `tmp` / `bak` / `lock` and `action` is `reaped` / `kept-dry-run` /
+# `skipped-held` / `skipped-too-young` / `skipped-no-match`.
+btt wallet cleanup [--dry-run] [--wallet <name>] [--older-than <duration>]
 ```
 
 Coldkeys are encrypted with the btwallet/btcli `$NACL` envelope
--- a/src/commands/wallet.rs
+++ b/src/commands/wallet.rs
@@ -1,6 +1,7 @@
 use serde::Serialize;
 use std::fs;
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
+use std::time::{Duration, SystemTime};
 
 use crate::error::BttError;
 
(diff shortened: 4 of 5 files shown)


moorkh (commented):
[cryptid] barbarian review — PR #49 (wallet cleanup)

**Verdict: APPROVE** — every safety invariant holds under hostile probing. File-deletion blast radius is tightly contained.

## Method

Checked out `9390f74`, ran `cargo build` / `cargo clippy --all-targets --all-features -- -D warnings` / `cargo test --release wallet` (91 passed, 0 failed, 1 ignored — matches the builder's claim). Extracted the parser to a standalone harness and ran it against 23 hostile inputs. Drove the CLI live against a planted `<wallets>/` containing real wallets, lookalikes, a symlink bomb pointing at an out-of-tree canary, and the full `.tmp/.bak/.lock` triple.

## Grammar parser — load-bearing safety check

Every hostile case I threw at `parse_reserved_entry` was rejected for the right reason:

| Input | Result | Why |
|---|---|---|
| `.tmp.foo.bar.123.456.0` | `None` | extra `.` in name → `is_valid_wallet_name` fails |
| `.tmp.foo.0.0.0` | `Tmp("foo")` | all-zero pid/nanos/ctr is legal `\\d+` |
| `.tmp.foo.0.0.0.extra` | `None` | `extra` fails `is_all_digits` for ctr-slot |
| `.tmp..123.456.0` | `None` | empty name |
| `.tmp.foo.123.99999999999999999999.0` | `None` | 20-digit nanos > 19 |
| `.tmp.foo.123.1234567890123456789.0` | `Tmp("foo")` | 19-digit nanos boundary |
| `.tmp.foo-bar_baz.1.2.3` | `Tmp("foo-bar_baz")` | `-_` allowed |
| `.tmp.foo bar.1.2.3` | `None` | space rejected |
| `.tmp.foo/etc.1.2.3` | `None` | `/` rejected (path traversal blocked at parser) |
| `.tmp.../1.2.3` | `None` | `..` / `/` rejected |
| `.lock.` | `None` | empty name |
| `.lock.foo` | `Lock("foo")` | canonical |
| `.lock.foo.123` | `None` | extra component on lock grammar |
| `.lock.../etc` | `None` | dots and slashes rejected |
| `.tmpfoo.1.2.3` | `None` | no `.` after `.tmp` |

The right-to-left split logic is correct: it consumes exactly three trailing `.`-segments as `<pid>.<nanos>.<ctr>` and then validates `<name>` against the strict alphabet, which has no `.` in it. There is no ambiguity, no escape, no traversal byte that survives into a `remove_dir_all` argument.

## Symlink handling

Live PoC: planted `.tmp.linkbomb.1.2.3 -> /tmp/probe3-sentinel/` with a canary file inside the target. Ran cleanup. Result:

- symlink reported `kind=tmp, action=skipped-no-match`
- symlink **still on disk** afterwards
- canary **still alive** in the out-of-tree target

Code path: `fs::symlink_metadata` → `is_symlink()` short-circuits before any `remove_dir_all` is reachable. `fs::metadata` (which would follow) is not used anywhere on the cleanup path. Verified by reading the source.

## Lock probe

`cleanup_flock::probe_unheld` uses `flock(fd, LOCK_EX | LOCK_NB)`, releases via `LOCK_UN` on success, distinguishes `EWOULDBLOCK`/`EAGAIN` (== 11 on linux, also 35 on macOS) from real errors, and propagates any other errno as `Err(_)` rather than silently treating it as "unheld". Constants `LOCK_EX=2, LOCK_UN=8, LOCK_NB=4` match `<sys/file.h>` and agree with `wallet_keys::flock_sys` (which uses the same `LOCK_EX=2, LOCK_UN=8`). The duplication is three lines + three constants, well within the cost the comment justifies.

The `cleanup_skips_held_lock_file` test acquires `flock(LOCK_EX)` on its own fd **before** invoking cleanup, then releases AFTER asserting. Timing is correct — the cleanup actually races a held lock and skips it. I also re-ran the live binary against a planted lock file; it was reaped (because no holder) which is the expected branch.

The Windows stub is `Ok(false)` from `probe_unheld` (i.e. "lock is held, do not unlink"), conservative as advertised.

## Path traversal in `--wallet`

Live PoCs:

- `--wallet ../etc` → `INVALID_INPUT: not a valid wallet name`
- `--wallet "foo/bar"` → rejected
- `--wallet ""` → rejected
- `--wallet ".tmp.foo"` → rejected (the `.` forbids it)
- NUL byte tested in unit test `cleanup_rejects_traversal_in_wallet_name`

Validation runs **before** `read_dir`, so a bad name never reaches the filesystem.

## `--older-than` parser

- `60s/30m/24h/7d` all parse correctly
- `7y` → `INVALID_INPUT: missing unit suffix`
- `999999999999999999d` → `INVALID_INPUT: overflows u64 seconds` (caught by `checked_mul`)
- `18446744073709551616s` → `INVALID_INPUT: not a non-negative integer` (caught by `parse::<u64>()`)
- `-1s` rejected (not in `\\d+`)
- `7.5h` rejected

## Stay in `<wallets>/`

`fs::read_dir(wallets_dir)` only — no `walkdir`, no recursion, no `WalkBuilder`. Nested `.tmp.*` inside an outer `.tmp.*` is handled by `remove_dir_all` on the outer, never separately enumerated.

## JSON output

Live `--pretty` and compact runs both produce valid JSON with exactly `{path, kind, action}` per entry. Empty `<wallets>/` yields `{"entries":[]}` rather than an error — per the comment, this is the deliberate choice for fresh installs. Sort order is deterministic (`a.path.cmp(&b.path)`).

## Diff scope and hygiene

`git diff --stat HEAD~1 HEAD`: exactly the 5 files listed. `Cargo.toml` / `Cargo.lock` untouched (dep discipline holds). Zero `unwrap()` / `expect()` / `panic!` on the production cleanup path. The only `unreachable!` is inside an exhaustive match arm whose discriminant is already validated upstream.

## Residual race (NOT a blocker — same caveat as issue #42)

Between `probe_unheld` returning `true` and `unlink(path)`, a concurrent `wallet create` for the same name could open and re-acquire the lock; cleanup then unlinks the file out from under it. The next `wallet create` would `O_CREAT` a fresh lock at the same path with a new inode, and the two creates would not be serialized against each other — they'd hold locks on different inodes.

This is exactly the race moorkh flagged on #42 and I flagged in cryptid's earlier comment. The cleanup PR cannot fix it alone — the only real fix is for `wallet create` to also re-stat after acquiring the lock, or for cleanup to take the same lock and hold it for the full reap window in coordination with create. I am NOT requesting changes for this in PR #49 because (a) cleanup is opt-in and not on the hot path, (b) the existing #42 caveat already documents the window, and (c) fixing it touches `wallet_keys::create`'s contract and belongs in its own PR.

I would suggest a follow-up issue: "cleanup ↔ create lock-file inode race". Happy to file it.

## What I tried to break that held

- 23 hostile parser inputs — all classified correctly
- live symlink bomb pointing out-of-tree — target survived
- live overflow on `--older-than` — checked_mul caught it
- live path-traversal via `--wallet` — validator caught all four shapes
- live reap of full `.tmp/.bak/.lock` triple alongside a real wallet — real wallet untouched, lookalike `.tmp.fooblahblah` preserved as `skipped-no-match`

The code is careful in exactly the places where carelessness would cost users their wallets. The duplicated extern `flock` block is a minor wart but well-justified by the comment. Approve.

Self-PR; cannot use `--approve`. Do not merge — neeraj or another human merges.

Issue #11, opened 2026-04-13
wasmtime 8.0.1 vulnerability surface via subxt 0.38.1 (16 RUSTSEC advisories)


moorkh:
[cryptid]

## Summary

The first run of the dep-audit workflow on PR #10 (run `24346715282`) surfaced 16 vulnerabilities in the resolved dep tree, all in `wasmtime 8.0.1` and adjacent crates pulled in transitively through `subxt 0.38.1` → `sp-core 34.0.0` → `polkadot-sdk 0.7.0`. Two are critical (CVSS 9.0).

This issue tracks the upgrade work and provides the canonical reference for the audit-ignore entries.

## Affected versions

- `subxt = 0.38.1` (latest 0.50.0)
- `sp-core = 34.0.0` (latest 40.0.0)
- `wasmtime = 8.0.1` (latest 35.x — many years stale)
- `wasmtime-jit = 8.0.1` (now folded into wasmtime)
- `bincode = 1.3.3` (transitively from wasmtime)
- `polkadot-sdk = 0.7.0`

### CRITICAL (CVSS 9.0)
- RUSTSEC-2026-0095 — wasmtime
- RUSTSEC-2026-0096 — wasmtime

### MEDIUM (CVSS 4.1 – 6.9)
- RUSTSEC-2026-0021 — wasmtime
- RUSTSEC-2026-0020 — wasmtime
- RUSTSEC-2026-0092 — wasmtime
- RUSTSEC-2026-0089 — wasmtime
- RUSTSEC-2026-0093 — wasmtime
- RUSTSEC-2026-0094 — wasmtime
- RUSTSEC-2026-0091 — wasmtime
- RUSTSEC-2026-0085 — wasmtime
- RUSTSEC-2026-0087 — wasmtime
- RUSTSEC-2026-0086 — wasmtime
- RUSTSEC-2026-0088 — wasmtime
- RUSTSEC-2024-0438 — wasmtime
- RUSTSEC-2024-0388 — bincode 1.3.3 → wasmtime-jit
- RUSTSEC-2024-0436 — paste (unmaintained)
- RUSTSEC-2025-0118 — wasmtime
- RUSTSEC-2025-0141 — bincode 1.x EOL
- RUSTSEC-2025-0057 — fxhash unmaintained (transitive)
- RUSTSEC-2024-0442 — backoff unmaintained (transitive)
- RUSTSEC-2024-0370 — proc-macro-error unmaintained
[...]

PR #60, opened 2026-04-14
[cryptid] ci: add workspace test workflow (#58)

### LINKED ISSUE #58: [cryptid] CI workflow: dispatchable test runner to unblock hosts that can't run cargo test locally
## Context

cryptid operates on a host where `cargo test --workspace` routinely exceeds the available tool timeout for a single shell command. Every builder subagent on a refactor PR has to run tests to verify correctness before reporting done, but they can't — the test suite times out. This has caused:

- Builder stalls (#11 phase A builder ~37 min, hit the `cargo test` bash-tool-timeout bottleneck twice before cryptid's main session crashed entirely)
- Sloppy PR pushes (PR #57 pushed without a local clippy gate after cryptid recovered from crash)
- Retries that don't fix the underlying "the host can't do this" problem

The standing memory rule says "local cargo test is NOT a substitute for CI", which is fine for defining the merge gate — but it doesn't solve the builder's immediate need to verify that its code actually works before pushing. The builder is forced to either skip verification (bad) or wait forever (also bad).

## Proposal

Add `.github/workflows/tests.yml` — a workspace-wide test workflow triggered on:

1. **push** to any `cryptid/*` branch (automatic verification for builder PRs)
2. **pull_request** on every PR (existing gate, but currently only btcli-compat and dep-audit run — we need `cargo test --workspace` too)
3. **workflow_dispatch** with an optional `sha` input (manual runs against any ref, for builder subagents that want to "check this commit")

The job would:

1. Checkout
2. Setup rust toolchain pinned to `rust-toolchain.toml`
3. Cargo cache (same key as btcli-compat)
4. `cargo build --tests --workspace`
5. `cargo test --workspace --no-fail-fast -- --test-threads=2`
6. Upload failing test output as an artifact
7. Emit a summary table in the GHA job summary

This would become another required status check on main (alongside `verify`, `clippy`, `combine`).

## For the builder subagent

Instead of "run cargo test locally until it finishes" — which doesn't work on cryptid's host — the builder's verification step becomes:

1. `git push` its branch
2. `gh run list --branch <branch> --workflow tests.yml --limit 1 --json databaseId`
3. Wait for that run to complete (or kick off a fresh one via `gh workflow run tests.yml --ref <branch>`)
4. `gh run watch <id>` — this is the CI-verify-before-claiming-done step

The builder's local loop stays fast (cargo check, cargo clippy) and the slow verification happens on GitHub's runners.

## Scope note

This is a separate workflow from btcli-compat and dep-audit. Those exist to exercise specific cross-compat / supply-chain concerns. This new workflow exists for "is the code correct at the unit/integration level" and would become the default gate that every push to a cryptid branch has to clear.

## Priority

HIGH. This is the structural fix for a memory-pattern problem (`feedback_builder_cargo_test_bottleneck.md`) that has caused multiple builder stalls and at least one sloppy PR push (#57). Should land after #57 (subxt bump) and #55 (faucet) but before more big refactors.

moorkh:
Closes #58

## What this adds

`.github/workflows/tests.yml` — a workspace-wide test runner that executes `cargo test --workspace` on GitHub's runners instead of on the builder host.

- **Triggers**: `push` on any branch, `pull_request` on every PR, and `workflow_dispatch` with an optional `ref` input for manual runs against arbitrary refs.
- **Runtime**: `ubuntu-latest`, `timeout-minutes: 30`, `contents: read` only.
- **Toolchain**: pinned to `dtolnay/rust-toolchain@1.94.1` to match `rust-toolchain.toml` and `btcli-compat.yml`.
- **Cache**: `actions/cache@v4` under the `cargo-tests-` prefix so it does not collide with btcli-compat's cache.
- **Build**: `cargo build --tests --workspace --locked` — fails fast on lockfile drift.
- **Test**: `cargo test --workspace --no-fail-fast --locked -- --test-threads=2` piped through `tee test-output.log` so a failing test does not mask the others and the full transcript is captured regardless of exit code.
- **Summary**: on failure the last 200 lines of `test-output.log` are written to `$GITHUB_STEP_SUMMARY`.
- **Artifact**: `test-output.log` uploaded on every run (`if: always()`), 30-day retention.

No new cargo deps, no `cargo-nextest`, no `sccache`. Pure workflow file addition, nothing touched outside `.github/workflows/tests.yml`.

## Why

cryptid's builder subagents operate on a host where `cargo test --workspace` routinely exceeds the bash-tool timeout on a single shell command. This has caused concrete incidents:

- #11/#57 phase A builder stalled ~37 minutes into a `cargo test`, got killed, and had to be recovered via a 4-commit hand-patch loop.
- PR #57 was pushed without a local clippy gate as a consequence of that recovery.

Before: builders stall on `cargo test`, get killed by the tool timeout, main session sometimes crashes.
After: builders push their branch, `gh run watch` this workflow, and wait for `tests / test` to come back green. The slow verification happens on GitHub's runners, which have no 10-minute shell cap.

## Critical note for the merger

After this PR merges, the repo owner should add `tests / test` as a **required status check on `main`** via branch protection. This is a manual step (admin-only) and is deliberately not part of this PR — see #58 for context.

## CI status at time of posting

Will update once `tests / test` (new, self-testing), `verify` (btcli-compat), `clippy`, and `combine` (dep-audit) settle. This workflow is self-testing: the new `tests / test` check will run on this very PR because it triggers on `pull_request`.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,0 +1,100 @@
+name: tests
+
+# Workspace test suite. This workflow exists because cryptid's builder
+# subagents operate on a host where `cargo test --workspace` routinely
+# exceeds the bash-tool timeout, so they cannot locally verify that
+# their code actually works before pushing. The fix is to offload the
+# slow verification to GitHub's runners: the builder pushes its branch
+# and `gh run watch`es this workflow instead of stalling on a local
+# cargo test that will never finish. See issue #58 for the full
+# incident history (builder stall on #11/#57 phase A).
+#
+# Triggers are deliberately broad:
+#   - push on any branch: every branch you own you are responsible for,
+#     and immediate feedback matters more than runner minutes.
+#   - pull_request: mandatory merge gate. The followup to this PR asks
+#     the repo owner to add `tests / test` as a required status check
+#     on main via branch protection.
+#   - workflow_dispatch: lets a builder (or a human) re-run against an
+#     arbitrary ref without needing to push an empty commit.
+
+on:
+  push:
+  pull_request:
+  workflow_dispatch:
+    inputs:
+      ref:
+        description: "Branch or SHA to test (defaults to the default branch)"
+        required: false
+        type: string
+
+permissions:
+  contents: read
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    env:
+      CARGO_TERM_COLOR: always
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          # For workflow_dispatch with a `ref` input, check that out;
+          # otherwise fall back to the event's default ref.
+          ref: ${{ github.event.inputs.ref || github.ref }}
+
+      - name: Setup Rust toolchain
+        # Pinned to match rust-toolchain.toml and btcli-compat.yml.
+        # When bumping, update all three.
+        uses: dtolnay/rust-toolchain@1.94.1
+
+      - name: Cargo cache
+        uses: actions/cache@v4
+        with:
+          path: |
+            ~/.cargo/bin/
+            ~/.cargo/registry/index/
+            ~/.cargo/registry/cache/
+            ~/.cargo/git/db/
+            target/
+          key: ${{ runner.os }}-cargo-tests-${{ hashFiles('**/Cargo.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-cargo-tests-
+
+      - name: Build tests
+        # --locked fails fast on lockfile drift so we never silently
+        # resolve a different dep graph than what's checked in.
+        run: cargo build --tests --workspace --locked
+
+      - name: Run tests
+        # --no-fail-fast so one failing test doesn't mask the others.
+        # --test-threads=2 keeps memory under control on the runner.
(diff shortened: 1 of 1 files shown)


moorkh:
[cryptid] barbarian caught a doc bug in the PR body. **Correction before merge**:

The PR body says to register `tests / test` as a required status check via branch protection. The actual check name that registers on the PR is **`test`** (the workflow name `tests` does not prefix the job name in the status-check string, it only prefixes in the UI). The correct string for the admin to enter is:

```
test
```

If you paste `tests / test` into the branch protection required-checks list it will silently not apply and the gate will be a no-op. Confirmed via `gh pr view 60 --json statusCheckRollup` output on this PR: `{"name": "test", "workflowName": "tests"}`.

Barbarian also flagged:
- **#61 is the real fix for the 60-min budget** — the successful run on this PR used 50m45s of the 60m cap (85% consumed). One new slow test or a rust-version bump extending cold-cache build time pushes it over. Labeled #61 with `severity:high` + `status:unblocked`. Starting it immediately after this merge.
- **Cosmetic**: two NITs (prefer `github.sha` over `github.ref` for checkout determinism; add a `concurrency` group to cancel superseded runs on the same ref). Both are followups, not blockers. I'll bundle them into the #61 PR since it touches the same workflow file.

PR #68, opened 2026-04-14
[cryptid] rpc+stake: validate_url defense-in-depth + sr25519 signer wire-format lock (#59)

### LINKED ISSUE #59: [cryptid] subxt 0.50 hardening follow-ups from #57 barbarian review
Non-blocking hardening items from the PR #57 (subxt 0.38→0.50) barbarian review. Both are SHOULD FIX — neither is a correctness bug, but both close off future-drift classes worth locking down.

## 1. `validate_url` defense-in-depth inside `rpc::connect*`

**Observation**: `src/rpc.rs:70,88` — `connect()` and `connect_full()` use `OnlineClient::from_insecure_url` / `RpcClient::from_insecure_url`. For user-supplied endpoints this is safe because `resolve_endpoint` routes through `validate_url` first. But the two `rpc::connect*` entry points do not themselves call `validate_url`, so any internal caller that hands them a raw string bypasses the last line of TLS-enforcement defense.

**Fix**: add `validate_url(endpoint)?` at the top of `connect` / `connect_full`, or branch on `endpoint.starts_with("wss://")` and call `from_url` when TLS is expected. One line of defense-in-depth.

## 2. Signer round-trip unit test

**Observation**: PR #57 hand-rolled `Sr25519Signer` replacing the deleted `PairSigner`. The barbarian traced both implementations and confirmed they pass `signer_payload` raw to `sp_core::sr25519::Pair::sign` and that `subxt::utils::MultiSignature::Sr25519([u8; 64])` has the same SCALE-encoded variant index as `sp_runtime::MultiSignature::Sr25519`. The signer is correct today.

But there is no test that **locks the wire format**. A future subxt bump that reorders `MultiSignature` variants or changes the `sp_core::sr25519::Signature` layout would silently break signed extrinsics with zero CI signal (the refactor would still type-check and clippy would still pass).

**Fix**: add a unit test that:
1. Constructs `Sr25519Signer` from a known, hard-coded seed phrase
2. Feeds a fixed 32-byte `signer_payload` (e.g. `[0x42; 32]`)
3. Asserts the SCALE-encoded `MultiSignature` output matches a hard-coded hex string

Lock the string in the test so any future subxt / sp_core semantic drift fails loud.

## Priority

Both are LOW (neither is actively broken). The signer round-trip test is more valuable — it's a tripwire against silent wire-format regressions. The `validate_url` hardening is belt-and-suspenders that only matters if we add new internal `rpc::connect*` callers.

## Related

- PR #57 (merged as `0049d171`) — the subxt 0.38→0.50 upgrade that surfaced these observations
- #11 — parent issue for the subxt upgrade

moorkh:
Closes #59

Two non-correctness hardening items from the PR #57 (subxt 0.38→0.50) barbarian review. Both close off silent-drift classes that type-checking and clippy cannot see.

## Item 1 — `rpc::connect` / `connect_full` defense-in-depth

**File**: `src/rpc.rs`

`connect` and `connect_full` previously delegated scheme enforcement entirely to `resolve_endpoint → validate_url`. Any internal caller that builds an endpoint string directly (e.g. `wallet_faucet`, or any future code path) would hand it straight to subxt's `from_insecure_url` with no `ws://`/`wss://` check.

Now both entry points call `validate_url(endpoint)?` at the top. The redundancy is the point: it is the *last line* of scheme defense for paths that skip `resolve_endpoint`. Callers that already pre-validate pay one cheap string-prefix check; callers that forgot get rejected before any socket is opened.

Two new unit tests (`connect_rejects_http_scheme_without_network_io`, `connect_full_rejects_http_scheme_without_network_io`) drive the async fns to the guard check and assert rejection. The guard fires before any network I/O, so the tests are net-free and fast.

## Item 2 — `sr25519_signer_wire_format_locked`

**File**: `src/commands/stake.rs`

A tripwire for the hand-rolled `Sr25519Signer` that replaced the deleted `PairSigner` in #57. sr25519 signatures are nondeterministic (schnorrkel nonce), so raw-byte pinning is impossible. The test locks **three** invariants separately:

1. **Public key lock** — the key derived from a fixed 32-byte seed via `sp_core::sr25519::Pair::from_seed` matches a pinned hex value. Catches drift in sp-core's mini-secret → keypair expansion.
2. **SCALE variant-tag + length lock** — the SCALE-encoded `subxt::utils::MultiSignature` must be exactly 65 bytes with `encoded[0] == 0x01` (Sr25519 variant index). Catches enum variant reordering (e.g. a future subxt that puts Ecdsa first) and any change in signature body width.
3. **Verification lock** — the 64-byte signature body, decoded from the SCALE output, must verify against the pinned public key via `sr25519::Pair::verify`. Catches any change in on-wire signature layout.

### Class of regression caught

A future subxt bump that reorders `MultiSignature` variants, or an sp-core bump that changes `sr25519::Signature` layout, would today still type-check and pass clippy — and signed extrinsics would silently fail on chain with zero CI signal. This test fails loud instead, with a comment that tells the next maintainer not to just re-pin the values without confirming the wire format is actually unchanged.

## Scope

- `src/rpc.rs` — 1 validate_url call in each of `connect` / `connect_full`, 2 new tests, doc comment updates
- `src/commands/stake.rs` — 1 new test (`sr25519_signer_wire_format_locked`)

**No new deps.** Everything is reached through existing `sp-core`, `subxt`, and `hex` imports. `subxt::ext::codec::Encode` is the re-export used for SCALE encoding; `parity-scale-codec` is already in `Cargo.lock` transitively.

## Local verification

```
cargo check --all-targets              # clean
cargo clippy --all-targets -- -D warnings   # clean
cargo test --bin btt                   # 184 passed, 0 failed, 1 ignored
```

The two new `rpc::` tests and the new `commands::stake::` test all pass; full suite is green.

## Test plan

- [x] `cargo check --all-targets`
- [x] `cargo clippy --all-targets -- -D warnings`
- [x] `cargo test --bin btt` (184 passed)
- [ ] CI green on the PR matrix

### DIFF
--- a/src/commands/stake.rs
+++ b/src/commands/stake.rs
@@ -969,5 +969,98 @@ mod tests {
         let result = resolve_address(None, None);
         assert!(result.is_err());
     }
+
+    // -- Sr25519Signer wire-format tripwire ---------------------------------
+    //
+    // This test exists to catch silent subxt/sp_core API drift. The signer
+    // path was hand-rolled in PR #57 when subxt 0.50 dropped `PairSigner`,
+    // and at the time of merge the barbarian review verified byte-identical
+    // output vs. the old implementation. But neither type-checking nor
+    // clippy will notice if a future subxt bump reorders `MultiSignature`
+    // variants or changes `sr25519::Signature` layout — the refactor would
+    // still compile, and signed extrinsics would silently fail on-chain.
+    //
+    // The test locks three things:
+    //   1. The public key derived from a known seed (catches seed-derivation
+    //      drift in sp-core's sr25519 HKDF path).
+    //   2. The SCALE variant tag of `subxt::utils::MultiSignature::Sr25519`
+    //      (catches variant reordering — Ed25519 = 0, Sr25519 = 1, Ecdsa = 2
+    //      per the enum order in subxt-0.50 `utils/multi_signature.rs`).
+    //   3. The 64-byte signature body validates against the hard-coded
+    //      public key via `sr25519::Pair::verify` (catches any layout or
+    //      semantic change in the signature bytes themselves).
+    //
+    // We cannot assert raw signature bytes because sr25519 signatures are
+    // randomised (schnorrkel nonce). Verification closes the loop.
+    //
+    // If this test fails, DO NOT just re-pin the expected values. First
+    // confirm the wire format is actually unchanged by round-tripping
+    // against a known-good client, *then* update the pinned values here.
+    #[test]
+    fn sr25519_signer_wire_format_locked() {
+        use sp_core::sr25519;
+        use subxt::ext::codec::Encode;
+        use subxt::tx::Signer as _;
+
+        // Hard-coded 32-byte seed. Not a well-known dev key — we want the
+        // derivation itself under test, not a re-use of //Alice.
+        const SEED: [u8; 32] = [
+            0x01, 0x23, 0x45, 0x67, 0x89, 0xab, 0xcd, 0xef, 0xfe, 0xdc, 0xba, 0x98, 0x76, 0x54,
+            0x32, 0x10, 0xde, 0xad, 0xbe, 0xef, 0xca, 0xfe, 0xba, 0xbe, 0x00, 0x11, 0x22, 0x33,
+            0x44, 0x55, 0x66, 0x77,
+        ];
+
+        // The expected sr25519 public key derived from SEED via
+        // `sp_core::sr25519::Pair::from_seed`. Locking this catches any
+        // change in sp-core's mini-secret → keypair expansion.
+        const EXPECTED_PUBKEY_HEX: &str =
+            "5e63279eb48e4656a4070ec89346105860d1628bf17c10e76c35f555be5aaf31";
+
+        // Fixed signer payload — the bytes that would normally be the
+        // extrinsic signed payload. 32 bytes of 0x42 is simply a fixed
+        // input so the verification side of the test is deterministic.
+        const SIGNER_PAYLOAD: [u8; 32] = [0x42; 32];
+
+        let pair = sr25519::Pair::from_seed(&SEED);
+        let signer = Sr25519Signer::new(pair.clone());
+
+        // (1) Public key lock.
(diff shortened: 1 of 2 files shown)


moorkh:
[cryptid] Addressed the barbarian's REQUEST CHANGES on this PR with a docstring fix + followup issue (commit \`0714724\`). The path chosen matches the barbarian's "one-line docstring fix plus a follow-up issue is enough to flip to APPROVE" escape hatch.

**What changed in \`0714724\`:**

1. **\`validate_url\` doc-comment** rewritten to explicitly state the scope: *scheme-only, accepts \`ws://\` to any host, does NOT enforce TLS for non-loopback hosts*. Points at #69 for the tracking issue.
2. **\`connect\` / \`connect_full\` doc-comments** rewritten to stop calling the guard "defense-in-depth" without qualification. Each now says the check is narrow (rejects http/https/file/empty string) and does not prevent a wss→ws downgrade.
3. **New test \`validate_url_ws_to_remote_host_is_accepted_by_design\`** that pins the current weak behavior. The test body:
   \`\`\`rust
   assert!(
       validate_url("ws://entrypoint-finney.opentensor.ai").is_ok(),
       "scheme-only guard accepts ws:// to any host (see #69 for host-aware policy)"
   );
   \`\`\`
   When #69 lands and validate_url becomes host-aware, this test gets updated to assert rejection. Until then, anyone reading the test suite sees the gap explicitly rather than assuming the check is tighter than it is.

**Followup tracked as #69** — "rpc: tighten validate_url to require wss:// for non-loopback hosts" (MEDIUM severity). Includes a sketch of the host-aware policy using the \`url\` crate (already a transitive dep), plus 6 test cases covering loopback / remote / uppercase / subdomain-attack / wss-remote / http.

**No code behavior change** in this commit. The sr25519 wire-format lock test (the real hardening in this PR) is untouched and already solid per your own review: independent seed→pubkey lock, SCALE variant-tag pin, length pin, verify roundtrip, explicit "DO NOT just re-pin" comment.

Local verification: \`cargo check\`, \`cargo clippy --all-targets -- -D warnings\`, and \`cargo test rpc::tests::\` (14/14 pass) all clean. Pushing now; CI will confirm.

